Issue #46, opened 2021-06-27
`module 'torch.distributed' has no attribute 'init_process_group'`


h-vetinari:
While trying to build https://github.com/conda-forge/allennlp-feedstock/pull/19, there were first issues à la:
```
from torch.distributed import ReduceOp
ImportError: cannot import name 'ReduceOp'
```

This seems to have been fixed in the most recent builds, but the allennlp test suite still fails with
```
module 'torch.distributed' has no attribute 'init_process_group'
```

The upstream AllenNLP devs say that things work with the builds from the pytorch channel.

dirkgr:
I am the upstream allennlp dev. I can confirm that this works with the pytorch channel builds.

isuruf:
That's expected given the default value. See https://github.com/pytorch/pytorch/blob/c3745dc5808aae7f3a6135de5e773286c46e4837/torch/distributed/__init__.py#L10-L15

dirkgr:
Can the setting be changed for this build of pytorch?

h-vetinari:
It seems the pytorch channel builds with `USE_DISTRIBUTED=1` on MacOS then? Would be good to have this for the conda-forge builds as well.

rgommers:
> It seems the pytorch channel builds with `USE_DISTRIBUTED=1` on MacOS then?

Indeed. Explanation for why it's not default in the repo on macOS: https://github.com/pytorch/builder/blob/master/conda/build_pytorch.sh#L133

h-vetinari:
> Indeed. Explanation for why it's not default in the repo on macOS: https://github.com/pytorch/builder/blob/master/conda/build_pytorch.sh#L133

Thanks. Since [`libuv`](https://github.com/conda-forge/libuv-feedstock) is available in conda-forge, there shouldn't be a fundamental roadblock to do the same in conda-forge?

Issue #32, opened 2021-01-13
Windows builds


h-vetinari:
There's CI for windows, but it fails. The only available builds are for `pytorch-cpu`, [v1.1.0](https://anaconda.org/conda-forge/pytorch-cpu/files) more than 1.5 years ago, nothing for the newer `pytorch` output yet.

Would be cool, as I'm trying to add builds for windows for https://github.com/conda-forge/allennlp-feedstock/pull/12

h-vetinari:
Just saw the work in #27, but even with `cuda_compiler_version=="None"`, the azure machine runs out of memory...

hmaarrfk:
i think we need somebody to build this locally then upload it. but licensing issues have stopped us from doing that with windows.

h-vetinari:
How do licencing issues come into this? 🤔 
(that wouldn't equally apply e.g. to https://github.com/conda-forge/qt-feedstock/)

hmaarrfk:
i'm not too sure. I think a ping to to the right person with a large windows machine might go a long way.

@isuruf any ideas?

isuruf:
Build fails with,
```
Building CXX object caffe2\torch\CMakeFiles\torch_python.dir\csrc\Device.cpp.obj
FAILED: caffe2/torch/CMakeFiles/torch_python.dir/csrc/Device.cpp.obj
[...]

PR #47, opened 2021-06-29
torch.distributed support for MacOS

IvanYashchuk:
Checklist

Fixes https://github.com/conda-forge/pytorch-cpu-feedstock/issues/46

torch.distributed is disabled by default on MacOS, to enable it we need to use `USE_DISTRIBUTED=1` when building PyTorch. The required dependency (`libuv`) is already listed in the `meta.yaml` file.


hmaarrfk:
Because this build fails alot, could you add a WIP and expliclitey ping us when you need help?

IvanYashchuk:
Yep, I marked this PR as a draft. I'll investigate the failures, but it's strange, basically, all I did is to add one flag to OSX build and rerender.

IvanYashchuk:
osx-arm64 fails because pkg-config doesn't work: `pkg-config.bin: Bad CPU type in executable`

windows failure: `conda_build.exceptions.DependencyNeedsBuildingError: Unsatisfiable dependencies for platform win-64: {'nccl=2', 'cudnn=undefined'}`

hmaarrfk:
nice finds. is there some issue we can refer to upstream regarding the proto if pinning

h-vetinari:
I commented on the pkg-config issue, but AFAICT this is a problem inherent in cross-compiling. For the moment, I would skip the distributed support on osx-arm64 and only build it for osx-x86.

h-vetinari:
> windows failure: `conda_build.exceptions.DependencyNeedsBuildingError: Unsatisfiable dependencies for platform win-64: {'nccl=2', 'cudnn=undefined'}`

Windows is currently broken, pending #32 resp. #27

IvanYashchuk:
PyTorch pins the protobuf that it uses in `third_party/protobuf` submodule. It's used with `BUILD_CUSTOM_PROTOBUF=ON` that was turned off recently here https://github.com/conda-forge/pytorch-cpu-feedstock/commit/5a7a3ebf9622d6c357c8f3bc4c6c8ccc2844245e. It seems to be needed only for building caffe2 and complaints about it are pretty common. Unfortunately, it's not obvious that the update to the recent protobuf version is simple to do.
From [Caffe2 FAQ](https://caffe2.ai/docs/faq.html#how-do-i-fix-error-messages-that-are-protobuf-related):
> Protobuf version mismatch is a common problem. Having different protobuf versions often leads to incompatible headers and libraries. ***Upgrading to the latest protobuf is not the solution.***

Here are related issues on the PyTorch repo: https://github.com/pytorch/pytorch/issues/59924, https://github.com/pytorch/pytorch/issues/20206.

hmaarrfk:
@isuruf  can you comment on the decision to set BUILD_CUSTOM_PROTOBUF=OFF ? It seems to help with the caffee builds and seems to be the "recommended" way to do things upstream.

recipe/meta.yaml
```diff
@@ -51,7 +50,8 @@ outputs:
         - git       # [not win]
         - m2-git    # [win]
         - ninja
-        - protobuf
+        # caffe2 doesn't compile with newer protobuf, so pinning it here
```
hmaarrfk, line 53: If caffe2 is vendoring protobuf, do it expect to communicate with other packages using the same protobuf version? If so, I don't think this will work.

If it is only using protobuf internally, and compiling it statically, we can likely safely remove it from our dependencies.

hmaarrfk:
@isuruf are we ok removing libprotobuf again?

hmaarrfk:
Are you able to test building windows CUDA builds/

IvanYashchuk:
Sorry, no, I was trying to debug it through CI. I should not continue that.
The original goal of this pull request - enabling distributed for MacOS - is reached and CI jobs pass. Along with that, I was trying to fix other failing builds:

- Linux times out at 6 hours,
- MacOS arm64 can't compile protobuf even though cross-compiling works on pytorch channel,
- Windows failures I don't understand yet.

I'm inclined to revert nonrelevant changes to get MacOS distributed support merged. It's fine to merge PRs with non-passing CI if it's not the current PR's fault, right?

As separate pull requests, I will try to get linux jobs to compile faster than 6h or set up ccache and restarting ci.
When I get a Windows machine, I will try to get Windows builds working.

isuruf:
> It's fine to merge PRs with non-passing CI if it's not the current PR's fault, right?

`osx-arm64` builds work just fine in master.

IvanYashchuk:
> > It's fine to merge PRs with non-passing CI if it's not the current PR's fault, right?
> 
> `osx-arm64` builds work just fine in master.

Yes, it will work in this PR as well with `USE_DISTRIBUTED=0` and protobuf from conda-forge, but protobuf needs to be pinned.

isuruf:
> but protobuf needs to be pinned.

What does this mean?

IvanYashchuk:
> > but protobuf needs to be pinned.
> 
> What does this mean?

Internal dependency of PyTorch, Caffe2, doesn't compile with the latest protobuf. It doesn't work with versions newer than 3.15.8.
See https://github.com/conda-forge/pytorch-cpu-feedstock/pull/47#issuecomment-872196725

isuruf:
Okay, then we should protobuf to the version needed.

isuruf:
Wait, what's the error exactly?

IvanYashchuk:
> error: This file was generated by a newer version of protoc which is incompatible with your Protocol Buffer headers. Please update your headers.
 
 CI jobs before I tried pinning: https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=344065&view=logs&j=15f8dc28-9f6f-5e3a-5b3c-2b169071e5be&t=1a4c7350-22aa-5d5b-9589-f36c9946ec39

hmaarrfk:
@isuruf i'm not sure why we need to pin protobuf. pytorch seems to be using it internally only and statically linking to it. 

If we use our own version, we will have people unable to install pytorch with their other software.

While I understand that we don't want people vendoring libraries, I think that pytorch is doing it to avoid conflicts with external libraries for an internal dependency of theirs. I don't think people are expected to interact with pytorch externally with protobufs.

Maybe I am wrong though.

isuruf:
Issue is that the arm64 builds are failing when using their custom version.

isuruf:
I haven't seen any errors about why we can't use the conda packaged newer protobuf. Can someone post the error when using the conda packaged newer protobuf?

isuruf:
That error can be easily fixed.
You need,
```
requirements:
  build:
    - protobuf {{ libprotobuf }}
```
to make sure the protoc in build matches that of the `libprotobuf` in `host`.

IvanYashchuk:
@hmaarrfk, this PR is ready for review.

Sorry for messing up with the CI trying to fix stuff not relevant to this PR. I'll do it later separately https://github.com/conda-forge/pytorch-cpu-feedstock/pull/47#issuecomment-880914019.

hmaarrfk:
can you remind me what activate in script does?

Do we need to rebuild all of the Linux matrix (cuda) that would require a lengthy secondary process (that i would mostly have to remember to start)

IvanYashchuk:
I removed "activate in script", because conda-forge linter was complaining about it   https://github.com/conda-forge/pytorch-cpu-feedstock/pull/47#issuecomment-870605370

hmaarrfk:
Lets keep the activate in script for now. I think it might be doing something with the cuda builds.

I'll wait until to night to make sure that the OSX builds pass then merge.

Maybe we can rerun some other migrations to ensure that when we start the long pytorch rebuild process, we aren't doing it multiple times.

hmaarrfk:
I think this is great.

I want to get a few other things merged before we rebuild the CUDA Linux stuff.
I don't think I'll be rebuilding all the cuda stuff just yet. But we can work toward it in the coming weeks.

Be careful with how you pin builds if you need to specify it in the recipes.

jakirkham:
JFYI we may want to add & drop a few CUDA versions at that point ( https://github.com/conda-forge/cudatoolkit-feedstock/issues/58 ) ( https://github.com/conda-forge/conda-forge-pinning-feedstock/pull/1708 )

hmaarrfk:
Thanks. the size of the cuda build matrix for this is a little unwieldy.

PR #78, opened 2022-01-04
Skip unwanted cuda 10.2, 11.0, and 11.1

hmaarrfk:
I figure I would leave this here to a vote, to decide if we wanted to skip unwanted cuda builds.

If so I can merge this in the libprotobuf PR.

Checklist


hmaarrfk:
@conda-forge/pytorch-cpu please vote by "approving" or "requesting changes"

recipe/meta.yaml
```diff
@@ -13,10 +13,13 @@ source:
...
 
 build:
-  number: 0
+  number: 1
   run_exports:
     - {{ pin_subpackage('pytorch', max_pin='x.x.x') }}
   skip: true  # [win]
+  skip: true  # [cuda_compiler_version=="10.2"]
+  skip: true  # [cuda_compiler_version=="11.0"]
+  skip: true  # [cuda_compiler_version=="11.1"]
```
jakirkham, line 22: Just so I understand, the goal is just to build with CUDA 11.2 and use CUDA Enhanced Compatibility for CUDA 11.2+?

Also we could [fuse these into a single `skip`]( https://github.com/conda-forge/mandrake-feedstock/blob/6b5ff9263188eaa9c316bdce3b418427bef9581c/recipe/meta.yaml#L14 )

```suggestion
  skip: true  # [cuda_compiler_version in ("10.2", "11.0", "11.1")]
```

hmaarrfk: yes you are correct.

isuruf:
What's the reason for this? Pain to build them manually?

hmaarrfk:
mostly a pain to build them manually.

hmaarrfk:

hmaarrfk:
@sodre, @benjaminrwilson do you feel strongly about dropping support for for cuda less than 11.2? Building pytorch many times is really a pain, and I am having trouble maintaining it on my own. 

Do let me know.

ngam:
> @sodre, @benjaminrwilson do you feel strongly about dropping support for for cuda less than 11.2? Building pytorch many times is really a pain, and I am having trouble maintaining it on my own.
> 
> Do let me know.

I am not a maintainer here and I have little vested interest, but I strongly vote to only build 11.2 (as you're proposing). I think if you want to be generous (and you don't really have to be in this case), you can keep the older archs and that's about it --- which you said you'd keep anyway. So I think there is no reason to ever build anything but 11.2 unless some people explicitly ask and give a strong reasoning for that. My 2c.

recipe/meta.yaml
```diff
@@ -13,10 +13,11 @@ source:
     - fix_dispatch_apply_auto.patch
 
 build:
-  number: 0
+  number: 1
   run_exports:
     - {{ pin_subpackage('pytorch', max_pin='x.x.x') }}
   skip: true  # [win]
+  skip: true  # [cuda_compiler_version in ("10.2", "11.0", "11.1")]
```
ngam, line 20: Could this be done more appropraitely if you simply put `cuda_compiler_version = 11.2` in the build config yaml?

ngam:
I also think we should just go ahead and build only cudatoolkit==11.2 going forward, it's been a month with nobody objecting.

hmaarrfk:
I guess people still use older drivers. So I'm going to close this until it becomes more of a pain to do.

ngam:
So I thought because of this part,

```
    export USE_CUDA=1
    export TORCH_CUDA_ARCH_LIST="3.5;5.0+PTX"
    if [[ ${cuda_compiler_version} == 9.0* ]]; then
        export TORCH_CUDA_ARCH_LIST="$TORCH_CUDA_ARCH_LIST;6.0;7.0"
    elif [[ ${cuda_compiler_version} == 9.2* ]]; then
        export TORCH_CUDA_ARCH_LIST="$TORCH_CUDA_ARCH_LIST;6.0;6.1;7.0"
    elif [[ ${cuda_compiler_version} == 10.* ]]; then
        export TORCH_CUDA_ARCH_LIST="$TORCH_CUDA_ARCH_LIST;6.0;6.1;7.0;7.5"
    elif [[ ${cuda_compiler_version} == 11.0* ]]; then
        export TORCH_CUDA_ARCH_LIST="$TORCH_CUDA_ARCH_LIST;6.0;6.1;7.0;7.5;8.0"
    elif [[ ${cuda_compiler_version} == 11.1 ]]; then
        export TORCH_CUDA_ARCH_LIST="$TORCH_CUDA_ARCH_LIST;6.0;6.1;7.0;7.5;8.0;8.6"
    elif [[ ${cuda_compiler_version} == 11.2 ]]; then
        export TORCH_CUDA_ARCH_LIST="$TORCH_CUDA_ARCH_LIST;6.0;6.1;7.0;7.5;8.0;8.6"
    else
        echo "unsupported cuda version. edit build_pytorch.sh"
        exit 1
    fi
```

we end up supporting all available arches from upstream, even deprecated ones, but I guess you're saying that some older systems cannot accept >11 and need something like 10.2, right? In that case, I completely misunderstood how this worked!

Issue #82, opened 2022-01-23
pytorch bombs if `"libblas=*=*accelerate"`


ngam:
- [x] I read [the conda-forge documentation](https://conda-forge.org/docs/user/introduction.html#how-can-i-install-packages-from-conda-forge) and could not find the solution for my problem there.

Issue:

pytorch bombs if `"libblas=*=*accelerate"`

reference: https://conda-forge.org/docs/maintainer/knowledge_base.html#blas

reference: https://github.com/pytorch/pytorch/blob/ea4d98388500496f04fd6a814b134c4242160267/cmake/Modules/FindBLAS.cmake#L122-L135

Not sure if this part of meta makes sense:

https://github.com/conda-forge/pytorch-cpu-feedstock/blob/5711a74b66df37cd4f5e879f111c25389052519f/recipe/meta.yaml#L69-L88

Here's the error when enforcing `"libblas=*=*accelerate"`:

```
Python 3.9.9 | packaged by conda-forge | (main, Dec 20 2021, 02:41:06) 
[Clang 11.1.0 ] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import torch 
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/ngam/.Mambaforge-MacOSX-arm64/envs/ptps/lib/python3.9/site-packages/torch/__init__.py", line 197, in <module>
    from torch._C import *  # noqa: F403
ImportError: dlopen(/Users/ngam/.Mambaforge-MacOSX-arm64/envs/ptps/lib/python3.9/site-packages/torch/_C.cpython-39-darwin.so, 0x0002): Library not loaded: @rpath/libopenblas.0.dylib
  Referenced from: /Users/ngam/.Mambaforge-MacOSX-arm64/envs/ptps/lib/python3.9/site-packages/torch/lib/libtorch_python.dylib
[...]

PR #83, opened 2022-01-23
fixing blas matrix?

ngam:


isuruf:
Please don't do this. Accelerate is a very experimental backend. Try instead to building with netlib variant which will give openblas by default and then users can use accelerate if they want to experiment.

ngam:
> Please don't do this. Accelerate is a very experimental backend. Try instead to building with netlib variant which will give openblas by default and then users can use accelerate if they want to experiment.

Yes, I agree. Do we just leave them undefined then?

isuruf:
> Yes, I agree. Do we just leave them undefined then?

Not sure. we have to convince CMake to use netlib instead of OpenBLAS.

ngam:
See 970bc53

ngam:
I will build this locally and test. I will post the results here :)

Ideally, I fully agree with the conda-forge knowlwedge base, unless we have to make a CMAKE choice, then I believe we should go for accelerate. Let me dig deeper to find how they define it for pytorch

ngam:
> > Yes, I agree. Do we just leave them undefined then?
> 
> Not sure. we have to convince CMake to use netlib instead of OpenBLAS.

@isuruf see issue #82

isuruf:
> I believe we should go for accelerate.

A huge NO. Accelerate is really buggy and ancient. It's on maintenance mode only. I would not recommend it for anything other than BLAS. Definitely no LAPACK functions.

ngam:
> > I believe we should go for accelerate.
> 
> A huge NO. Accelerate is really buggy and ancient. It's on maintenance mode only. I would not recommend it for anything other than BLAS. Definitely no LAPACK functions.

No, no, the idea is only to replace OpenBLAS with Accelerate. Everything else remains the same, including LAPACK.

ngam:
I am going to close this for now; I have no idea what's causing these issues (it is saying libprotobuf is undefined). We can revisit at a later stage

hmaarrfk:
I'm not sure why we haven't updated. the truth is that manual uploads are really straining on conda-forge core. I'm trying not to ping them too much. but there are many PRs in flight for pytorch ATM

ngam:
Yeah, I am definitely getting out of this mess now! I don't want to add to the pressure especially this seems to be a losing fight with a lot of moving pieces that I personally don't have use for at the moment. These are really, really complicated builds. Very frustrating!!!

ngam:
Do we generally follow the upstream pinning? For example for torchvision below in the table, we need to pin to v1.11.1... which we didin't build
 
<!DOCTYPE html>

torch | torchvision | python
-- | -- | --
main / nightly | main / nightly | >=3.7, <=3.9
1.10.1 | 0.11.2 | >=3.6, <=3.9
1.10.0 | 0.11.1 | >=3.6, <=3.9
1.9.1 | 0.10.1 | >=3.6, <=3.9
1.9.0 | 0.10.0 | >=3.6, <=3.9
1.8.2 | 0.9.2 | >=3.6, <=3.9
1.8.1 | 0.9.1 | >=3.6, <=3.9
1.8.0 | 0.9.0 | >=3.6, <=3.9
1.7.1 | 0.8.2 | >=3.6, <=3.9
1.7.0 | 0.8.1 | >=3.6, <=3.8


<br class="Apple-interchange-newline">

hmaarrfk:
hmm oops. yeah. we do try.

ngam:
```
source:
  # for local testing use a tarball including submodules
  git_url: https://github.com/pytorch/pytorch.git
  git_tag: {{ commit }}
  patches:
    - fix_dispatch_apply_auto.patch
```

Are we not using a tar here because of the 3rd party stuff imbedded as submodules? I can volunteer to tidy that up --- I following a tar ball approach is more efficient and clearly imo

hmaarrfk:
How do you keep all the submodules up to date? Feel free to take a stab at it.

ngam:
> > I believe we should go for accelerate.
> 
> A huge NO. Accelerate is really buggy and ancient. It's on maintenance mode only. I would not recommend it for anything other than BLAS. Definitely no LAPACK functions.

Can you elaborate? It seems supported as far as I can tell: https://developer.apple.com/documentation/accelerate

rgommers:
> No, no, the idea is only to replace OpenBLAS with Accelerate. Everything else remains the same, including LAPACK.

I think you are missing a point here. Both OpenBLAS and Accelerate provide BLAS _and_ LAPACK.



> Can you elaborate? It seems supported as far as I can tell: https://developer.apple.com/documentation/accelerate

See for example https://github.com/scipy/scipy/wiki/Dropping-support-for-Accelerate. Some bugs have been fixed, but Accelerate is still stuck on the ancient LAPACK 3.2.1.

ngam:
> I think you are missing a point here. Both OpenBLAS and Accelerate provide BLAS _and_ LAPACK

Yes, sorry, I was mislead by my work on mxnet where it was printing them out differently (or maybe borrowing them from different place?). See this comment: https://github.com/conda-forge/pytorch-cpu-feedstock/pull/84#issuecomment-1019815924

ngam:
> Both OpenBLAS and Accelerate provide BLAS _and_ LAPACK

Okay, I was under the impression we are able to do this very easily, but maybe I am completely wrong: https://github.com/scipy/scipy/wiki/Dropping-support-for-Accelerate#keep-using-accelerate-blas-but-shadow-lapack

ngam:
Let me double check how it is doing it in my mxnet build, please have a look if you have time: https://github.com/conda-forge/staged-recipes/pull/17620

ngam:
@rgommers, I saw your name around this discussion about Accelerate; from your expert opinion, what should we do in general here (in conda-forge) if the only two options are OpenBLAS and Accelerate? Do I get your stance correctly below?

- Use OpenBLAS (and ignore Accelerate) if LAPACK is important (utilized enough in a given package)
- Use Accelerate (and ignore OpenBLAS) if LAPACK is unimportant (not really utilized in a given package)

Some reports indicate at least double the performance when using Accelerate, but we would want to be careful such that we are not messing up people's environments even for that gain in performance.  I also opened an issue upstream at PyTorch to see what they think and if we can run some tests to establish a preference.

isuruf:
@ngam, you are complicating this way more than it needs to be. We should just build with `netlib` variant and let the user decide between OpenBLAS and Accelerate.

ngam:
> @ngam, you are complicating this way more than it needs to be. We should just build with netlib variant and let the user decide between OpenBLAS and Accelerate.

Okay, thanks @isuruf. I can try that in the future once we have other more pressing issues sorted here. But I am still uncertain we can pull it off; I need to read the notes in the conda-forge knowledge base and see if they necessarily conflict with PyTorch/MXNet/etc. and their CMake processes.

ulupo:
> We should just build with netlib variant and let the user decide between OpenBLAS and Accelerate.

... and is the idea that the user would opt in Accelerate by as per https://conda-forge.org/docs/maintainer/knowledge_base.html#switching-blas-implementation ?

ngam:
> > We should just build with netlib variant and let the user decide between OpenBLAS and Accelerate.
> 
> ... and is the idea that the user would opt in Accelerate as per [standard instructions for switching](https://conda-forge.org/docs/maintainer/knowledge_base.html#switching-blas-implementation)?

Yes, and in this case, PyTorch does indeed fully support Accelerate and so it should work fine.

hmaarrfk:
excited to see a PR! In the past, we've used the test phase, and there chose different implementations of blas with selectors based on the python version.

Exciting!

ngam:
Yes, will do soon @hmaarrfk. I just want to dig deeper a little bit, I opened an issue upstream to discuss and their response is still unclear (though they are responsive and helpful).

Btw, I think we may be doing a better job packaging this than they do --- I think they drop MKL in their cuda builds among other things, but they have no reason to, I think. So thank you for all your work!!

PR #92, opened 2022-02-07
Test

isuruf:
Checklist


.cirun.yml
```diff
@@ -0,0 +1,8 @@
+runners:
+  - name: long-runner
+    cloud: aws
+    instance_type: c5d.4xlarge
+    machine_image: ami-0048c368c80e3ee99
+    preemptible: false
+    workflow: .github/workflows/conda-build.yml
+    count: 2
```
isuruf, line 8: @aktech, I tried a few different instance_types and they all fail with `no space left on device` while pulling the docker image. Any ideas?

aktech: Yes, that's expected with docker images. The default size of instance is less than 10 GB, which is not really enough.  

I can potentially add the setting/configuration in the `.cirun.yml` to increase disk size but that depends on disk type etc and there are lot of permutations for the same.

So, I would suggest to create a custom AMI (say on top of Ubuntu 20.04) with disk size something like say 30 GB or more and use that AMI ID.

wolfv:
Only using 13% of CPU 😢 

Screenshot 2022-02-08 at 15 49 27

isuruf:
That's probably because of https://github.com/conda-forge/conda-forge-ci-setup-feedstock/blob/master/recipe/run_conda_forge_build_setup_linux#L27

wolfv:
I thought it's because of `pip` as a build system not supporting multi-core compilation? Doesn't the line you linked just set a default?

isuruf:
> I thought it's because of pip as a build system not supporting multi-core compilation? 

pytorch uses pip, but it calls cmake under the hood.

> Doesn't the line you linked just set a default?

Yes, the default of 2 is not good

isuruf:
It is if the user doesn't set it. We set it there, because usually the value set up by conda-build is wrong. What happens is that the VM that CI gives us will run out of memory or some other resource. We can make an exception for github_actions just like we do for drone.

wolfv:
I don't remember having these issues when running `build_locally.py` on large machines (e.g. on the OVH machine that I used for TensorFlow).

isuruf:
> Should we stop this for now?

sure

> I don't remember having these issues when running build_locally.py on large machines (e.g. on the OVH machine that I used for TensorFlow).

No clue. It might be that tensorflow recipe is not configured to use `CPU_COUNT` variable.

isuruf:
@wolfv, shall we start another build to see if the CPU_COUNT fix is working now?

wolfv:
yes!

wolfv:
maybe we can remove some configs here? We have a 256 core limit

isuruf:
> maybe we can remove some configs here? We have a 256 core limit

I'm not sure I understand. core limit in what?

wolfv:
We (I) have a limit of 256 vCPUs on AWS. 

In general, I just think it might be enough to run 2 or so test configs in parallel.

wolfv:
Anyways, let's just let it run. I am excited :)

isuruf:
@aktech, is there a way to limit the number of runners run in parallel?

wolfv:
I was thinking about removing some configurations from this test build. But it doesn't matter that much for now.

aktech:
> @aktech, is there a way to limit the number of runners run in parallel?

Not from the Cirun interface at the moment, but since I am working on retry mechanism, the failed runner creations would retry and will create runners eventually, when the running jobs get finished. If that helps?

isuruf:
@wolfv, started a build.

@aktech, yeah, that might work. Thanks for the info. I just realized there's a `count` parameter, but I'm not sure what it does and doesn't seem to be documented.

aktech:
> Thanks for the info. I just realized there's a count parameter, but I'm not sure what it does and doesn't seem to be documented.

It's an old function: https://github.com/AktechLabs/cirun-docs/pull/8 It has been removed now as it's not required anymore. (apologies for the fact that it didn't followed any deprecation cycle or documentation update)

So, this is due to the previous webhook, I have been using called as `check_run` which is a single webhook event for the whole workflow, that's why there was a need to creating multiple VMs based on the count, but since now I am using the new `workflow_job` webhook, the `count` parameter is not required any more, as each job has a separate webhook event.

wolfv:
Hmm, AWS still shows only ~15% utilization. I don't know enough about AWS to judge wether this is normal / expected or not and I can't log in to the machines to check with `htop`.

aktech:
Hey @wolfv you can easily log onto machine to check with htop via: https://github.com/marketplace/actions/debugging-with-tmate

wolfv:
still seems really slow. on the other hadn IDK what the expected build time is ...

wolfv:
@aktech do you know if we should expect close to 100% CPU usage on AWS wiht a multicore build?

wolfv:
@isuruf these builds are still using MAX_JOBS=2 ( https://github.com/conda-forge/pytorch-cpu-feedstock/runs/5181076142?check_suite_focus=true#step:3:2075 )

isuruf:
> To create a local Git repository instead, add Git 2.18 or higher to the PATH

Looks like we need an image with Git 2.18 or higher for actions/checkout to get a git repo.

aktech:
@isuruf @wolfv Do you need any help with image creation here or with CI in general?

isuruf:
@aktech, yes. Help with image creation is greatly appreciated.

aktech:
I'll be happy to create the image, should I create it on my AWS account or can I have access to the AWS account which you guys are using?

isuruf:
> I'll be happy to create the image, should I create it on my AWS account or can I have access to the AWS account which you guys are using?

Thanks for the offer. I think it's best to create it on your account as I still need to figure out if I can give permissions to you.

aktech:
Hey @isuruf I created the image with latest Git (2.35.1) on top of the Ubuntu 18.04 with Docker (ami-0048c368c80e3ee99).

Here are the job runs on my fork: https://github.com/aktech/pytorch-cpu-feedstock/actions/runs/1952646838

The AMI ID is `ami-0febc229056b2de48` in `eu-west-1` (Ireland region). I have made it public, I believe you can use it as well.

aktech:
Both jobs [completed successfully.](https://github.com/aktech/pytorch-cpu-feedstock/actions/runs/1952646838)

.cirun.yml
```diff
@@ -0,0 +1,8 @@
+runners:
+  - name: long-runner
+    cloud: aws
+    instance_type: c5d.4xlarge
+    machine_image: ami-0582cf60af51d3113
```
aktech, line 5:
```suggestion
    machine_image: ami-0febc229056b2de48
```

.cirun.yml
```diff
@@ -0,0 +1,8 @@
...
+    preemptible: false
+    workflow: .github/workflows/conda-build.yml
+    count: 2
```
aktech, line 8:
```suggestion
```

Tobias-Fischer:
Can we close here @isuruf?

Issue #71, opened 2021-10-28
track features should be on the bare pytorch package


hmaarrfk:
https://github.com/conda-forge/pytorch-cpu-feedstock/blob/b8824904342b8ea142d297668bf0aa32790091a6/recipe/meta.yaml#L149

xref https://github.com/conda-forge/tensorflow-feedstock/pull/157

hmaarrfk:
While this is true of the current deployment, up until 1.1.0 pytorch did have the two `pytorch-cpu` package name
https://anaconda.org/pytorch/pytorch-cpu/files

We decided to keep the historical packaging names.

We can remove them later, but I think this current issue should be resolved.

hmaarrfk:
Is it worth finally biting the bullet and implementing this change? For what its worth, it would put it in line with https://github.com/conda-forge/tensorflow-feedstock/blob/main/recipe/meta.yaml#L127

PR #105, opened 2022-05-24
 [ci skip] [skip ci] [cf admin skip] ***NO_CI*** add track_feature to cpu builds to deprioritize them

hmaarrfk:
Closes https://github.com/conda-forge/pytorch-cpu-feedstock/issues/71

cc: @adamjstewart @ashnair1 

I'm asking the bot to skip these builds so that we can do a repo-patch instead. I'm worried that the increased build number will cause some solver confusion. Honestly, i also can't dedicate resources to building tis for an entire day or two this week.

repodata patch: https://github.com/conda-forge/conda-forge-repodata-patches-feedstock/pull/265

Checklist

### DIFF
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -12,7 +12,7 @@ source:
     - fix_dispatch_apply_auto.patch
 
 build:
-  number: 1
+  number: 2
   skip: true  # [win]
 
 outputs:
@@ -23,6 +23,8 @@ outputs:
       detect_binary_files_with_prefix: false
       run_exports:
         - {{ pin_subpackage('pytorch', max_pin='x.x') }}
+      track_features:
+        - pytorch-cpu                                      # [cuda_compiler_version == "None"]
     script: build_pytorch.sh  # [not win]
     script: bld_pytorch.bat   # [win]
     requirements:
@@ -109,6 +111,8 @@ outputs:
         - typing_extensions
         # Need ninja to load C++ extensions
         - ninja
+        # avoid that people without GPUs needlessly download ~0.5-1GB
+        - __cuda  # [cuda_compiler_version != "None"]
       run_constrained:
         # These constraints ensure conflict between pytorch and
         # pytorch-cpu 1.1 which we built before conda-forge had GPU infrastructure


rgommers:
Isn't `track_features` legacy and shouldn't be used? IIRC `pytorch` on the `pytorch` channel moved away from it and to a mutex because `track_features` didn't work for `mamba`.

hmaarrfk:
@rgommers it is deprecated, I think it is mostly a hack that we used in tensorflow to weigh down the CPU implementation in the case that the user has a GPU: https://github.com/conda-forge/tensorflow-feedstock/blob/main/recipe/meta.yaml#L230

ngam:
I support this to keep the logic consistent across tensorflow and pytorch (and in the future other packages like them).

Having said that, I am much more in favor of asking people to select the build variant themselves, so my personal solution to problem around the `__cuda` condition and GPU discovery is to actually *always* install the cpu variant unless the cuda variant is explicitly requested via `*cuda*` in the call. Having seen how many issues have come up recently, I am wary of the burden of maintaining this and issues around it in the future (it's just too confusing with many moving parts).

However, I do think this PR will improve the status quo, and so I support it.

adamjstewart:
> Having said that, I am much more in favor of asking people to select the build variant themselves

To add some context, in [TorchGeo](https://github.com/microsoft/torchgeo), we have a single `environment.yml` in which we list the dependencies of the project. We would love to always specify `pytorch-gpu` as a dependency so that a GPU-capable version of PyTorch is installed even if the user doesn't have a GPU, but there is no version of `pytorch-gpu` available with Windows support, so we have to rely on `pytorch` instead. I think your point makes sense for command-line installs, but doesn't make sense for `environment.yml` installs. We _could_ have two `environment.yml` (the only difference being `pytorch-gpu` vs. `pytorch-cpu`, but this becomes yet another file in which we need to list the project's many dependencies and keep it up-to-date.

ngam:
> > Having said that, I am much more in favor of asking people to select the build variant themselves
> 
> To add some context, in [TorchGeo](https://github.com/microsoft/torchgeo), we have a single `environment.yml` in which we list the dependencies of the project. We would love to always specify `pytorch-gpu` as a dependency so that a GPU-capable version of PyTorch is installed even if the user doesn't have a GPU, but there is no version of `pytorch-gpu` available with Windows support, so we have to rely on `pytorch` instead. I think your point makes sense for command-line installs, but doesn't make sense for `environment.yml` installs. We _could_ have two `environment.yml` (the only difference being `pytorch-gpu` vs. `pytorch-cpu`, but this becomes yet another file in which we need to list the project's many dependencies and keep it up-to-date.

You're correct that I was merely thinking about the command-line case only... and I never take windows into account... we actually don't have a windows build available here. 

I think there is a way to specify more things about the architecture in your yaml file (but I am not sure) and I am not sure if conda supports the pip-esque sys.platform control flow https://peps.python.org/pep-0496/#examples

ngam:
Also, there is no rush to get this done now. How about adding it in the next release? (Less than a month away, it seems, so instead of skipping ci etc we can just wait a bit more)

adamjstewart:
> I think there is a way to specify more things about the architecture in your yaml file (but I am not sure) and I am not sure if conda supports the pip-esque sys.platform control flow https://peps.python.org/pep-0496/#examples

From what I can tell this isn't possible in [`environment.yml`](https://docs.conda.io/projects/conda/en/latest/user-guide/tasks/manage-channels.html), only in a build recipe [`meta.yaml`](https://docs.conda.io/projects/conda-build/en/latest/resources/define-metadata.html). Also, the `meta.yaml` has a lot of limitations. You can't add a dependency specific to a single version of Python without converting the generic pure-Python universal build to a platform/Python-version-specific build. Overall, conda has very poor support for conditional dependencies compared to pip/spack.

hmaarrfk:
@adamjstewart while it would be "convenient" to install `pytorch-gpu` always, it is somewhat misleading to the end user. They don't have a GPU, why should they pay for the extra storage cost (it is likely close to a 2GB download for many GPU packages all said and done). Generally speaking, I am of the opinion that installing `pytorch-gpu`  should fail if the user doesn't have a GPU. We've discussed this in an other thread which resulted in the addition of documentation by ngam. 

At the end of the day, the proposed PR would solve your use case:
- Add `pytorch` to your dependencies
1. Users with a GPU would prioritize the GPU package. Thus have GPU capabilities.
2. Users without a GPU would find that the gpu variant, while prioritized, is incompatible, and thus rejected by the solver. The would only have CPU capabilities.

> Windows support

I don't think we have any windows builds here. Even a CPU PR would be welcome.

ngam:
> 2\. Users without a GPU would find that the gpu variant, while prioritized, is incompatible, and thus rejected by the solver. The would only have CPU capabilities.
> 
> >

Don't you need to add `__cuda` in the run section for this like in tensorflow tho?

hmaarrfk:
> > 2. Users without a GPU would find that the gpu variant, while prioritized, is incompatible, and thus rejected by the solver. The would only have CPU capabilities.
> > >
> 
> Don't you need to add `__cuda` in the run section for this like in tensorflow tho?

I don't think so since `pytorch` has the dependency on nccl.

ngam:
> I don't think so since `pytorch` has the dependency on nccl.

tensorflow has always had nccl dep, but we still added `__cuda`. Looking at nccl, it doesn't look it has a mechanism to prevent `__cuda`: https://github.com/conda-forge/nccl-feedstock/blob/main/recipe/meta.yaml

ngam:
<details>

```

ngam@m142:~$ conda info

     active environment : None
            shell level : 0
       user config file : /home/ngam/.condarc
 populated config files : /home/ngam/.Mambaforge-Linux-x86_64/.condarc
                          /home/ngam/.condarc
          conda version : 4.12.0
    conda-build version : not installed
         python version : 3.9.12.final.0
       virtual packages : __linux=5.4.0=0
                          __glibc=2.31=0
                          __unix=0=0
                          __archspec=1=x86_64
       base environment : /home/ngam/.Mambaforge-Linux-x86_64  (writable)
      conda av data dir : /home/ngam/.Mambaforge-Linux-x86_64/etc/conda
  conda av metadata url : None
           channel URLs : https://conda.anaconda.org/conda-forge/linux-64
                          https://conda.anaconda.org/conda-forge/noarch
          package cache : /home/ngam/.Mambaforge-Linux-x86_64/pkgs
                          /home/ngam/.conda/pkgs
       envs directories : /home/ngam/.Mambaforge-Linux-x86_64/envs
                          /home/ngam/.conda/envs
               platform : linux-64
             user-agent : conda/4.12.0 requests/2.27.1 CPython/3.9.12 Linux/5.4.0-113-generic ubuntu/20.04.4 glibc/2.31
                UID:GID : 1000:1000
             netrc file : None
           offline mode : False


ngam@m142:~$ mamba create -n test_pt pytorch=*=*cuda*

                  __    __    __    __
                 /  \  /  \  /  \  /  \
                /    \/    \/    \/    \
███████████████/  /██/  /██/  /██/  /████████████████████████
              /  / \   / \   / \   / \  \____
             /  /   \_/   \_/   \_/   \    o \__,
            / _/                       \_____/  `
            |/
        ███╗   ███╗ █████╗ ███╗   ███╗██████╗  █████╗
        ████╗ ████║██╔══██╗████╗ ████║██╔══██╗██╔══██╗
        ██╔████╔██║███████║██╔████╔██║██████╔╝███████║
        ██║╚██╔╝██║██╔══██║██║╚██╔╝██║██╔══██╗██╔══██║
        ██║ ╚═╝ ██║██║  ██║██║ ╚═╝ ██║██████╔╝██║  ██║
        ╚═╝     ╚═╝╚═╝  ╚═╝╚═╝     ╚═╝╚═════╝ ╚═╝  ╚═╝

        mamba (0.23.3) supported by @QuantStack

        GitHub:  https://github.com/mamba-org/mamba
        Twitter: https://twitter.com/QuantStack

█████████████████████████████████████████████████████████████


Looking for: ['pytorch=[build=*cuda*]']

conda-forge/linux-64                                        Using cache
conda-forge/noarch                                          Using cache
Transaction

  Prefix: /home/ngam/.Mambaforge-Linux-x86_64/envs/test_pt

  Updating specs:

   - pytorch=*[build=*cuda*]


  Package                  Version  Build                   Channel                    Size
─────────────────────────────────────────────────────────────────────────────────────────────
  Install:
─────────────────────────────────────────────────────────────────────────────────────────────

  + _libgcc_mutex              0.1  conda_forge             conda-forge/linux-64     Cached
  + _openmp_mutex              4.5  2_kmp_llvm              conda-forge/linux-64        6kB
  + bzip2                    1.0.8  h7f98852_4              conda-forge/linux-64     Cached
  + ca-certificates    2022.5.18.1  ha878542_0              conda-forge/linux-64     Cached
  + cffi                    1.15.0  py310h0fdd8cc_0         conda-forge/linux-64      444kB
  + cudatoolkit             11.7.0  hd8887f6_10             conda-forge/linux-64      872MB
  + cudnn                 8.2.1.32  h86fa8c9_0              conda-forge/linux-64      707MB
  + ld_impl_linux-64        2.36.1  hea4e1c9_2              conda-forge/linux-64     Cached
  + libblas                  3.9.0  14_linux64_mkl          conda-forge/linux-64       13kB
  + libcblas                 3.9.0  14_linux64_mkl          conda-forge/linux-64       13kB
  + libffi                   3.4.2  h7f98852_5              conda-forge/linux-64     Cached
  + libgcc-ng               12.1.0  h8d9b700_16             conda-forge/linux-64     Cached
  + liblapack                3.9.0  14_linux64_mkl          conda-forge/linux-64       13kB
  + libnsl                   2.0.0  h7f98852_0              conda-forge/linux-64     Cached
  + libprotobuf             3.20.1  h6239696_0              conda-forge/linux-64        3MB
  + libstdcxx-ng            12.1.0  ha89aaad_16             conda-forge/linux-64     Cached
  + libuuid                 2.32.1  h7f98852_1000           conda-forge/linux-64     Cached
  + libzlib                 1.2.11  h166bdaf_1014           conda-forge/linux-64     Cached
  + llvm-openmp             14.0.4  he0ac6c6_0              conda-forge/linux-64        6MB
  + magma                    2.5.4  h6103c52_2              conda-forge/linux-64       93MB
  + mkl                   2022.1.0  h84fe81f_915            conda-forge/linux-64      209MB
  + nccl                 2.12.12.1  h0800d71_0              conda-forge/linux-64      142MB
  + ncurses                    6.3  h27087fc_1              conda-forge/linux-64     Cached
  + ninja                   1.11.0  h924138e_0              conda-forge/linux-64        3MB
  + numpy                   1.22.4  py310h4ef5377_0         conda-forge/linux-64        7MB
  + openssl                  3.0.3  h166bdaf_0              conda-forge/linux-64     Cached
  + pip                     22.1.1  pyhd8ed1ab_0            conda-forge/noarch       Cached
  + pycparser                 2.21  pyhd8ed1ab_0            conda-forge/noarch       Cached
  + python                  3.10.4  h2660328_0_cpython      conda-forge/linux-64       30MB
  + python_abi                3.10  2_cp310                 conda-forge/linux-64        4kB
  + pytorch                 1.11.0  cuda112py310h51fe464_1  conda-forge/linux-64      510MB
  + readline                   8.1  h46c0cb4_0              conda-forge/linux-64     Cached
  + setuptools              62.3.2  py310hff52083_0         conda-forge/linux-64        1MB
  + sleef                    3.5.1  h9b69904_2              conda-forge/linux-64        2MB
  + sqlite                  3.38.5  h4ff8645_0              conda-forge/linux-64     Cached
  + tbb                   2021.5.0  h924138e_1              conda-forge/linux-64        2MB
  + tk                      8.6.12  h27826a3_0              conda-forge/linux-64     Cached
  + typing_extensions        4.2.0  pyha770c72_1            conda-forge/noarch         28kB
  + tzdata                   2022a  h191b570_0              conda-forge/noarch       Cached
  + wheel                   0.37.1  pyhd8ed1ab_0            conda-forge/noarch       Cached
  + xz                       5.2.5  h516909a_1              conda-forge/linux-64     Cached
  + zlib                    1.2.11  h166bdaf_1014           conda-forge/linux-64     Cached

  Summary:

  Install: 42 packages

  Total download: 3GB

─────────────────────────────────────────────────────────────────────────────────────────────

Confirm changes: [Y/n] 

```

</details>

ngam:
Btw, I think your logic is sounder than our approach tbh; `__cuda` and `track_features` should definitely go into the actual GPU packages (e.g. cudatoolkit, nccl, cudnn, etc.) and *not* here and tensorflow.

leofang:
I don't think it's correct. There are desire to be able to install CUDA packages on systems without GPU (or CUDA driver), for example when compiling packages in a CI. If you set __cuda to run, these use cases would fail.

hmaarrfk:
@leofang we somewhat had this discussion before.

For the special cases you mention, we decided that it was appropriate to ask users to set the conda appropriate environment variable.

If you are interested in reading through the conversation, i can find it in the tensorflow PRs.

The conversation ended with  ngam writing documentation outlining to users how they would install tensorflow GPU on systems that didn't have GPUs.

ngam:
> @leofang we somewhat had this discussion before.
> 
> For the special cases you mention, we decided that it was appropriate to ask users to set the conda appropriate environment variable.
> 
> If you are interested in reading through the conversation, i can find it in the tensorflow PRs.
> 
> The conversation ended with ngam writing documentation outlining to users how they would install tensorflow GPU on systems that didn't have GPUs.

Yes, and I was the one who was arguing your point @leofang lol (but they convinced me this is for the best). 

The point in these issues is to formalize this standard and make it more efficient. The cuda ru constraint should be put in the packages which actually facilitate the cuda stuff (e.g. cudnn, nccl, cudatoolkit) instead of having to do that in every package downstream. Otherwise, we will end up with inconsistencies and problems imo

ngam:
(We can run this issue by the core team again to make it more formal, but the only way to adopt it ecosystem-wide is to put the constraint in the cuda packages; I have no power personally, I am just trying to make things easier as we push forward)

leofang:
I don't remember when/where the conversation happened, sorry. The most recent one I could think of is this PR https://github.com/conda-forge/cudatoolkit-feedstock/pull/76 which (IMHO correctly) fixed `__cuda` in `run_constraint`.

ngam:
@leofang let's discuss the details in https://github.com/conda-forge/cudatoolkit-feedstock/issues/78 so not to change the convo here

leofang:
If you just need a mechanism to deprioritize a certain build kind, couldn't you just use the build-number trick that already works for MPI? https://conda-forge.org/docs/maintainer/knowledge_base.html#preferring-a-provider-usually-nompi

leofang:
Sure I'll copy my comment there.

hmaarrfk:
I might have used the word "we" in the sense of "the collective" rather than "those that are present in the current discussion"

hmaarrfk:
@ngam do you mind reviewing to make sure i got the pinning correct?

ngam:
Yes, I will have a look soon.

ngam (approved):
This looks good to me. It does add the track_features as well as the __cuda condition. Everything else remains the same as desired. Bumping the build number to move forward. All good! Thanks!!!

hmaarrfk:
I somewhat don't want to bump the build number, and resort to repo-data patching. I should check that the hash is the same. I'm also attempting to skip the builds on azure when it is merge


- [ ] Same has for CPU builds

ngam:
> I somewhat don't want to bump the build number, and resort to repo-data patching. I should check that the hash is the same. I'm also attempting to skip the builds on azure when it is merge
> 
> * [ ]  Same has for CPU builds

If you're worried about this disrupting things, it is totally okay to wait until we have 1.12 (in a few weeks). Then we can do a full build cycle, testing and all, including bumping pytorch pinning. There is no urgency in doing this now imo.

Overall, if you're worried about this, I think a repo patch may be even more worrisome... my personal inclination is to leave the past in the past, so people have ample time to figure out the new stricter regime (i.e. things will keep working as is for <=1.11)

What do you think?

hmaarrfk:
i somewhat tend to agree. but i'm kinda lazy at the moment and don't really have that much CPU on hand. we could just remember to do this for 1.12. that seems ok.

ngam:
> i somewhat tend to agree. but i'm kinda lazy at the moment and don't really have that much CPU on hand. we could just remember to do this for 1.12. that seems ok.

Let's track this and other tasks for 1.12 in #107

hmaarrfk:
@leofang do you still have strong oppositions to this? If no, I would like to fold it into: https://github.com/conda-forge/pytorch-cpu-feedstock/pull/116 and start a rebuild.

ngam:
> @leofang do you still have strong oppositions to this? If no, I would like to fold it into: #116 and start a rebuild.

fwiw, I think leofang's opposition was more about adding `__cuda` in cudnn, nccl, and cudatoolkit and not specific to pytorch (if I understood correctly)

hmaarrfk:
FYI: @leofang https://mamba.readthedocs.io/en/latest/advanced_usage/package_resolution.html

leofang:
Thanks Mark, sorry I am still digesting the weekend backlog. Right, I don't have issues for CUDA-downstream packages to track `__cuda`. The mamba page is a good resource indeed (one that we should also add to conda-forge...)

leofang:
Looking at this a little more, I find the recipe is a bit confusing. I'd expect that we build two outputs (`pytorch-cpu`, `pytorch-gpu`) separately, and have the meta package `pytorch` sitting on top. The current recipe looks to me like a workaround based on a then-CPU-only recipe. Though if it works fine and is well maintained, I think it's enough 🙂 

If I were asked to review, the remaining question would be why not just add `__cuda` to `run_constrained`, given that `track_features` is already used here to prioritize CPU builds.

hmaarrfk:
`run_constrained` would not stop you from installing the GPU version with a CPU only system.

Given the higher priority to the GPU builds, they would always get installed on a CPU. For what its worth, I decided to implement your idea of the build numbers in https://github.com/conda-forge/pytorch-cpu-feedstock/pull/116 which barring any objects from maintainers of pytorch or conda-forge core, i was hoping to merge in a few days.

leofang:
Thanks for pointer Mark!

PR #120, opened 2022-06-23
Include in 1.12 release: Duplicate deps in pytorch meta.yml

apatsekin:
Hey guys, just deleting duplicates in meta.yml. Feel free to drop the PR if it's somehow on purpose.


hmaarrfk:
Thank you for the PR. Given the difficulty in managing builds for pytorch, i've just flagged this for inclusion in 1.12 and we will try to ensure we include this cleanup then.

PR #137, opened 2022-09-25
migrate to libprotobuf321

ngam:
Checklist
* [ ] Used a [personal fork of the feedstock to propose changes](https://conda-forge.org/docs/maintainer/updating_pkgs.html#forking-and-pull-requests)
* [ ] Bumped the build number (if the version is unchanged)
* [ ] Reset the build number to `0` (if the version changed)
* [ ] [Re-rendered]( https://conda-forge.org/docs/maintainer/updating_pkgs.html#rerendering-feedstocks ) with the latest `conda-smithy` (Use the phrase <code>@<space/>conda-forge-admin, please rerender</code> in a comment in this PR for automated rerendering)
* [ ] Ensured the license file is being packaged.

PR #138, opened 2022-10-04
Drop cuda 10.2 and 11.0

hmaarrfk:
in https://github.com/conda-forge/pytorch-cpu-feedstock/pull/137

we contemplated dropping Cuda 10.2 and 11.0 and 11.1 due to maintainer burden.

This leaves 11.2 that was released in December 2020.

I'm looking for feedback from others interested in Pytorch at conda-forge.

I'm proposing leaving this PR open until the end of November 2022 when hopefully we can make a decision.

cc: @xhochy @mikemhenry


hmaarrfk:
cc: @ngam

mikemhenry:
Eh, I think it depends on what the vision is for `conda-forge`. It is an entirely volunteer org (AFAIK) with finite resources, so we can't support a very comprehensive/non EOL'ed matrix of cuda versions. The space I'm most familiar with is HPC and I don't know of any national compute centers that don't have drivers that will work for at least cuda 11.2. I can't think of any academic code that I work with that can't build against cuda 11.2. I imagin most people have some macros to build code either using the 10.2 API or the new 11.X API. I sometimes think because people can compile 10.2 cuda and 11.2, they want and expect both but I'm not sure if we need it. 

Do we have any packages that have pytorch as a dependency that failed the cuda 11.2 migration? I think if all the deps for this package have 11.2 builds, then we should (technically) be fine to proceed. Dropping support for the older cuda versions will also have the effect of dropping older cuda versions for any environment with `pytorch` in the stack--which could make some users upset.

Personally, I am in favor of it. I'd love to see a page on https://conda-forge.org/docs/ that has a matrix/table of the numpy, openssl, cuda, python versions (and whatever else makes sense) we support building against and the dates we will drop/add new versions.

rgommers:
Seems very reasonable to me as well. If anything, supporting newer CUDA versions is more important than keeping old ones around. 11.2 is a good baseline, and then 11.4 (/11.6) should arrive at some point.

As a perhaps relevant data point, the PyTorch team would like to drop 10.2 on PyPI but isn't able too, because the 11.X wheels are >1 GB and PyPI won't accept that.

h-vetinari:
> and then 11.4 (/11.6) should arrive at some point.

Conda-forge already supports >=11.2, the only difference is that we don't need to rebuild for every minor version anymore due to the "enhanced [ABI] compatibility" since 11.2. :)

hmaarrfk:
> Do we have any packages that have pytorch as a dependency that failed the cuda 11.2 migration? 

I believe you helped alot with different migration. Please ping anybody you think might be interested in commenting.

I'm not too interested in making this a sweeping conda forge change. Let's keep things small.

We can vote by leaving a review. Green for yes. Request changes for no

h-vetinari:
> Once we have only 11.2+ here, then the big two (the other being tensorflow) will be only 11.2+ and the entire ecosystem should likely follow suit for compatibility purposes.

Not every CUDA enabled package depends on pytorch or tensorflow.

For the moment, I think it's mostly still a feedstock-specific question. For pytorch and tensorflow in particular it's hard enough already to get _any_ packages, so 11.2+ is _much_ better IMO than blocking on full CUDA support.

But for the ecosystem, this discussion should happen in https://github.com/conda-forge/conda-forge.github.io. Based on the previous [discussion](https://github.com/conda-forge/conda-forge-pinning-feedstock/pull/1708) (>1yo!), the appetite was not there yet.

Note that I'm not against this in principle, but l suggest to separate the ecosystem discussion from the pytorch one.

xhochy:
As I have been tagged here: I don't have any good CUDA experience at the moment, thus I would omit myself for now from the discussion.

hmaarrfk:
> As I have been tagged here: I don't have any good CUDA experience at the moment, thus I would omit myself for now from the discussion.

I guess you just helped a lot with tensorflow and other builds in the past!

jakirkham:
The main challenge has been getting cloud providers to upgrade in a timely manner. For example here's an issue with Kaggle still on CUDA 11.0 ( https://github.com/Kaggle/docker-python/issues/1098 ). The image they are using comes from Google. So it's honestly out of Kaggle's hands.

To be clear, I don't really expect anyone here to solve this issue (or frankly even support these older CUDA versions longer). Completely understand all of that is a pain and would like to drop them myself. That said, I do worry that there are users that will get cut off from support unfortunately due to things largely out of their control and just wanted to relay that concern here (with one example for context)

mikemhenry:
CC @kkraus14

jakirkham:
cc @jaimergp

jaimergp:
From the Colab perspective, we have merged today a new version that *should* allow for any cudatoolkit compatible with the present drivers.

Colab's `nvidia-smi` reports:

```
Tue Oct  4 21:38:49 2022       
+-----------------------------------------------------------------------------+
| NVIDIA-SMI 460.32.03    Driver Version: 460.32.03    CUDA Version: 11.2     |
|-------------------------------+----------------------+----------------------+
| GPU  Name        Persistence-M| Bus-Id        Disp.A | Volatile Uncorr. ECC |
| Fan  Temp  Perf  Pwr:Usage/Cap|         Memory-Usage | GPU-Util  Compute M. |
|                               |                      |               MIG M. |
|===============================+======================+======================|
|   0  Tesla T4            Off  | 00000000:00:04.0 Off |                    0 |
| N/A   46C    P8     9W /  70W |      0MiB / 15109MiB |      0%      Default |
|                               |                      |                  N/A |
+-------------------------------+----------------------+----------------------+
                                                                               
+-----------------------------------------------------------------------------+
| Processes:                                                                  |
|  GPU   GI   CI        PID   Type   Process name                  GPU Memory |
|        ID   ID                                                   Usage      |
|=============================================================================|
|  No running processes found                                                 |
+-----------------------------------------------------------------------------+
```

So I think we are fine there.

If the concern is leaving some Kaggle users unsupported in newer versions, I'd say they can resort to the `pytorch` channel. Also, it looks like tensorflow 2.10 is only building for CUDA 11.2 already. Previous versions are available on conda-forge for CUDA 11.0 if needed.

This is not the first time we talk about dropping CUDA versions, but at some point we need to let go I guess 😬 That said, I am not as involved in the CUDA ecosystem anymore so I don't know if I should vote for something I am not too invested. Can I vote neutral but if there's a tie it can be understood as a "Yes, let's remove"?

jaimergp:
> I'm proposing leaving this PR open until the end of November 2022 when hopefully we can make a decision.

I'd say we can also make an announcement and request for comments in a more visible place at conda-forge.org, the Gitter rooms, Twitter, etc.

hmaarrfk:
I just did gitter and https://github.com/conda-forge/conda-forge-pinning-feedstock/issues/3491 i can't reach twitter, but I could tweat at conda-forge?

jakirkham:
In terms of making an announcement, we can add a PR like this one ( https://github.com/conda-forge/conda-forge.github.io/pull/1815 ) and provide a brief summary of what we plan to do and when, which will show up on [this page]( https://conda-forge.org/docs/user/announcements.html ) once merged. Users subscribed to the feed there will get an update.

jaimergp:
> I just did gitter and [conda-forge/conda-forge-pinning-feedstock#3491](https://github.com/conda-forge/conda-forge-pinning-feedstock/issues/3491) i can't reach twitter, but I could tweat at conda-forge?

Yes, I am not sure who has access to the Twitter account but let's do that: tweet it and condaforge acct will amplify.

Btw, John and I were told that Kaggle users are expecting a Docker image update soon including 11.3, so the 11.0 drop is a bit less concerning.

jakirkham:
Also have been looking into other CSPs internally with a colleague and it seem most have moved to CUDA 11.2 since we last looked, which is encouraging. Again that shouldn't affect the decision here either way, but it is a useful data point.

hmaarrfk:
Given the lack of any objections so far and the new release of Pytorch, I think we will likely merge this on by Nov 21st.

h-vetinari:
> Given the lack of any objections so far and the new release of Pytorch, I think we will likely merge this on by Nov 21st.

This (and the end of November mentioned elsewhere) has now come to pass :)

Issue #108, opened 2022-05-28
Splitting this package in managable chunks


hmaarrfk:
### Comment:

This package currently requires more than 16 builds to be build manually to ensure that it completes in time on the CIs.

# Step 1: No more git clone
rgommers identified that one portion of the build process that takes time is cloning the repository. In my experience, cloning the 1.5GB repo can take up to 10 min on my powerful local machine, but I feel like it can take much longer on the CIs.

To avoid cloning, we will have to list out all the submodule manually, or make the conda-forge installable dependencies.

I mostly got this working using a recursive script which should help us keep it maintained: https://github.com/conda-forge/pytorch-cpu-feedstock/pull/109

# Option 1: Split off Dependencies:

| Dependency     | linux  | mac    | win | GPU Aware | PR                                                       | system deps                                                                                    |
|----------------|--------|--------|-----|-----------|----------------------------------------------------------|------------------------------------------------------------------------------------------------|
| pybind11       |        |        |     | no        | https://github.com/conda-forge/pybind11-feedstock        | USE_SYSTEM_PYBIND11                                                                            |
| cub            |        |        |     | no        | https://github.com/conda-forge/cub-feedstock             |                                                                                                |
| eigen          |        |        |     | no        | https://github.com/conda-forge/eigen-feedstock           | [USE_SYSTEM_EIGEN_INSTALL](https://github.com/pytorch/pytorch/blob/master/CMakeLists.txt#L265) |
| googletest     |        |        |     | no        | will not package                                         |                                                                                                |
| benchmark      |        |        |     | no        | https://github.com/conda-forge/benchmark-feedstock       |                                                                                                |
| protobuf       |        |        |     | no        | https://github.com/conda-forge/libprotobuf-feedstock     |                                                                                                |
| ios-cmake      |        |        |     |           | not needed since we don't target ios                     |                                                                                                |
| NNPACK         | yes    | yes    |     | no        | https://github.com/conda-forge/staged-recipes/pull/19103 |                                                                                                |
| gloo           | yes    | yes   |     | yes     | https://github.com/conda-forge/staged-recipes/pull/19103 | USE_SYSTEM_GLOO                                                                                |
| pthreadpool    | yes    | yes    |     | no        | https://github.com/conda-forge/staged-recipes/pull/19103 | USE_SYSTEM_PTHREADPOOL                                                                         |
[...]

PR #154, opened 2022-12-17
Testing GH Actions instead of Azure

hadim:
Checklist


hmaarrfk:
Can you skip most builds? maybe

```
skip: true  # [py!=310]
```

hadim:
@hmaarrfk here is a test using GH Actions instead of Azure. As suggested (https://github.com/conda-forge/pytorch-cpu-feedstock/issues/108#issuecomment-1356355748), I had to comment out the line blocking GH Actions for conda-forge in `conda-smithy` during `regenerate`.

I guess the goal here is to check how long the CI takes. I don't think it's available for free, but updating the current CI to larger runners is supposed to one-click-away from a `conda-forge` GH org admin.

hmaarrfk:
I understand the test. But we don't need to build all of these just to test. At this stage, we can limit the build matrix until most things are debugged.

hadim:
`skip: true  # [py!=310]` generates no CI.

hmaarrfk:
i think maybe the syntax changed and they don't want us to use 310 anymore

```
skip: true  # [py38]
skip: true  # [py39]
skip: true  # [py311
```

hmaarrfk:
I guess lets see how far the linux builds go

hadim:
The regular GH runners seems similar to the Azure ones (2 CPUs, 6h limit).

I guess the build time will be similar.

Only large runners will make a difference: https://docs.github.com/en/actions/using-github-hosted-runners/using-larger-runners#machine-specs-for-larger-runners

But those are not free even for public repos. And it sounds like paying for it (even for quick testing) will require a lot of bureaucracy xD

hadim:
> Are you aware of larger runners for osx btw?

For now, it seems to be only for linux and windows. I guess it's a matter of time before osx is proposed.

hmaarrfk:
Personally, if we are going to pay money from a limited pool to get this built on linux, i would appreciate some help in unbundling things, so that we only have to use the larger runners once, for the GPU builds.

PR #159, opened 2023-01-24
Add llvm-openmp to host deps on osx

Tobias-Fischer:
Checklist
* [x] Used a [personal fork of the feedstock to propose changes](https://conda-forge.org/docs/maintainer/updating_pkgs.html#forking-and-pull-requests)
* [x] Bumped the build number (if the version is unchanged)
* [ ] Reset the build number to `0` (if the version changed)
* [x] [Re-rendered]( https://conda-forge.org/docs/maintainer/updating_pkgs.html#rerendering-feedstocks ) with the latest `conda-smithy` (Use the phrase <code>@<space/>conda-forge-admin, please rerender</code> in a comment in this PR for automated rerendering)
* [ ] Ensured the license file is being packaged.

llvm-openmp should be pinned to 14. However, for some reason the llvm-openmp build version (not host version) is used for the version pin at the moment. This should resolve this issue.

Issue originally encountered in https://github.com/conda-forge/norse-feedstock/pull/11 where:
```
package pytorch-1.13.0-cpu_py310h09eaf1a_0 requires llvm-openmp >=15.0.6, but none of the providers can be installed
```

Both osx-64 and osx-arm64 have implicitly already pulled in llvm-openmp in the host deps in previous builds.

PR #162, opened 2023-02-03
deps cleanup

### LINKED ISSUE #160: add llvm-openmp to host
### Solution to issue cannot be found in the documentation.

- [X] I checked the documentation.

### Issue

xref https://github.com/conda-forge/pytorch-cpu-feedstock/pull/159

As per the PR

### Installed packages

```shell
n/a
```

### Environment info

```shell
n/a
```

### LINKED ISSUE #161: Do we really need to run-depend on ninja?
ninja as a run-dep was added in https://github.com/conda-forge/pytorch-cpu-feedstock/commit/d1b3b1f04369c59e50fd0cd7d491f5ef7ca3c64e by @isuruf.

The comment says "Need ninja to load C++ extensions", but shouldn't that be on the consuming feedstock to have ninja available, or is this necessary even just to use pytorch?

I ask because having ninja present in the host-env in https://github.com/conda-forge/torchtext-feedstock/pull/7 picks up the wrong ninja during cross-compilation (not the one from the build env with the right CPU type), and it seems a bit gratuitous to me to run-depend on a build tool like that.

ngam:
Fix #160 
Fix #161 
Close #159 

Checklist

### DIFF
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -1,5 +1,5 @@
 {% set version = "1.13.1" %}
-{% set number = 0 %}
+{% set number = 1 %}
 
 {% if cuda_compiler_version != "None" %}
 {% set number = number + 200 %}
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -106,8 +106,6 @@ outputs:
         - setuptools
         - cffi
         - typing_extensions
-        # Need ninja to load C++ extensions
-        - ninja
         # avoid that people without GPUs needlessly download ~0.5-1GB
         - __cuda  # [cuda_compiler_version != "None"]
       run_constrained:
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -43,7 +43,6 @@ outputs:
         - python                                 # [build_platform != target_platform]
         - cross-python_{{ target_platform }}     # [build_platform != target_platform]
         - numpy                                  # [build_platform != target_platform]
-        - cffi                                   # [build_platform != target_platform]
         - sysroot_linux-64  2.17  # [linux64]
         - {{ compiler('c') }}
         - {{ compiler('cxx') }}
@@ -81,7 +80,6 @@ outputs:
         - requests
         - future
         - six
-        - cffi
         - mkl-devel {{ mkl }}   # [x86]
         - libcblas * *_mkl      # [x86]
         - libcblas              # [not x86]
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -87,6 +87,8 @@ outputs:
         - libcblas              # [not x86]
         - liblapack             # [not x86]
         - openblas              # [not x86]
+        - libgomp   # [linux]
+        - llvm-openmp    # [osx]
         - libprotobuf
         - sleef
         - typing
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -102,9 +102,6 @@ outputs:
         - {{ pin_compatible('magma', max_pin='x.x.x') }}      # [cuda_compiler_version != "None"]
         # other requirements
         - python
-        - {{ pin_compatible('numpy') }}
-        - setuptools
-        - cffi
         - typing_extensions
         # avoid that people without GPUs needlessly download ~0.5-1GB
         - __cuda  # [cuda_compiler_version != "None"]
--- a/.azure-pipelines/azure-pipelines-linux.yml
+++ b/.azure-pipelines/azure-pipelines-linux.yml
@@ -8,46 +8,46 @@ jobs:
     vmImage: ubuntu-latest
   strategy:
     matrix:
-      linux_64_c_compiler_version10cuda_compiler_version11.2cxx_compiler_version10numpy1.20python3.8.____cpython:
-        CONFIG: linux_64_c_compiler_version10cuda_compiler_version11.2cxx_compiler_version10numpy1.20python3.8.____cpython
+      linux_64_c_compiler_version10cuda_compiler_version11.2cudnn8cxx_compiler_version10numpy1.20python3.8.____cpython:
+        CONFIG: linux_64_c_compiler_version10cuda_compiler_version11.2cudnn8cxx_compiler_version10numpy1.20python3.8.____cpython
         UPLOAD_PACKAGES: 'True'
         DOCKER_IMAGE: quay.io/condaforge/linux-anvil-cuda:11.2
-        SHORT_CONFIG: linux_64_c_compiler_version10cuda_c_hc644ec66fe
-      linux_64_c_compiler_version10cuda_compiler_version11.2cxx_compiler_version10numpy1.20python3.9.____cpython:
-        CONFIG: linux_64_c_compiler_version10cuda_compiler_version11.2cxx_compiler_version10numpy1.20python3.9.____cpython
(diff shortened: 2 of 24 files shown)


recipe/meta.yaml
```diff
@@ -104,8 +106,6 @@ outputs:
         - setuptools
```
hmaarrfk, line 106: can we get rid of numpy and setuptools too now?

hmaarrfk:
How did you get around the issues with OSX SDK?

h-vetinari:
> I am not sure if there are any real downsides with setting the SDKs higher (not the deployment target) and I don't know why we have it really low by default.

As always: compatibility for old systems, though of course part of it is also inertia in changing something at all. With [`_LIBCPP_DISABLE_AVAILABILITY`](https://conda-forge.org/docs/maintainer/knowledge_base.html#newer-c-features-with-old-sdk) & per-feedstock opt-out from the baseline 10.9, most things can be solved as necessary. And for anything beyond that, there's this discussion: https://github.com/conda-forge/conda-forge.github.io/issues/1844

hmaarrfk:
While you are using the newer SDK, you are still targetting the older deployment targets unless you specify one more variable.

https://conda-forge.org/docs/maintainer/knowledge_base.html#requiring-newer-macos-sdks

hmaarrfk:
see https://github.com/conda-forge/conda-forge-ci-setup-feedstock/pull/228#issuecomment-1416810041

recipe/meta.yaml
hmaarrfk: yes very likely.

isuruf:
> I am just not sure why we set the SDK so low if we can set it higher with no downsides.

There are downsides. Most build systems don't understand the difference between the two (SDK version vs deployment target).
Even for a prominent package like python, only python 3.9+ understand the difference whereas for python 3.8, the build system assumes that the depolyment target and the SDK version are the same.

recipe/meta.yaml
```diff
@@ -100,12 +102,7 @@ outputs:
         - {{ pin_compatible('magma', max_pin='x.x.x') }}      # [cuda_compiler_version != "None"]
         # other requirements
         - python
-        - {{ pin_compatible('numpy') }}
```
hmaarrfk, line 103: if we cannot remove numpy from the build time dependneices, this should be moved to a run_constraint

hmaarrfk: Their package doesn't seem to depend on at runtime
image

Their recipe seems to look like:
<details>

```yaml
{% set build_variant = environ.get('PYTORCH_BUILD_VARIANT', 'cuda') %}
{% set cross_compile_arm64 = environ.get('CROSS_COMPILE_ARM64', 0) %}

package:
  name: pytorch
  version: "{{ environ.get('PYTORCH_BUILD_VERSION') }}"

source:
  path: "{{ environ.get('PYTORCH_GITHUB_ROOT_DIR') }}"

requirements:
  build:
    - cmake
    - {{ compiler('c') }} # [win]
    - pkg-config # [unix]
    - libuv # [unix]

  host:
    - python
    - setuptools
    - pyyaml
    {% if cross_compile_arm64 == 0 %}
    - mkl-include # [x86_64]
    - mkl=2020.2 # [x86_64 and (not win or py <= 39)]
    - mkl=2021.4 # [x86_64 and win and py >= 310]
    {% endif %}
    - typing_extensions
    - ninja
    - libuv # [win]
    - numpy=1.19 # [py <= 39]
    - numpy=1.21.5 # [py >= 310]
    - openssl=1.1.1l # [py >= 310 and linux]
{{ environ.get('PYTORCH_LLVM_PACKAGE', '    - llvmdev=9') }}
{{ environ.get('MAGMA_PACKAGE', '') }}

  run:
    - python
    {% if cross_compile_arm64 == 0 %}
    - mkl >=2018 # [x86_64]
    {% endif %}
    - libuv # [win]
    - intel-openmp # [win]
    - typing_extensions
    {% if cross_compile_arm64 == 0 %}
    - blas * mkl
    {% endif %}
    - pytorch-mutex 1.0 {{ build_variant }}  # [not osx ]
{{ environ.get('CONDA_CUDATOOLKIT_CONSTRAINT', '') }}

  {% if build_variant == 'cpu' %}
  run_constrained:
    - cpuonly
  {% elif not osx %}
  run_constrained:
     - cpuonly <0
  {% endif %}

build:
  number: {{ environ.get('PYTORCH_BUILD_NUMBER', '1') }}
  detect_binary_files_with_prefix: False
  string: "{{ environ.get('PYTORCH_BUILD_STRING') }}"
  script_env:
    - BUILD_SPLIT_CUDA
    - CUDA_VERSION
    - CUDNN_VERSION
    - CONDA_CUDATOOLKIT_CONSTRAINT
    - USE_CUDA
    - CMAKE_ARGS
    - EXTRA_CAFFE2_CMAKE_FLAGS
    - DEVELOPER_DIR
    - DEBUG
    - USE_FBGEMM
    - USE_GLOO_WITH_OPENSSL # [unix]
    - USE_SCCACHE # [win]
    - USE_DISTRIBUTED # [unix]
    - CMAKE_OSX_ARCHITECTURES # [unix]
    - USE_MKLDNN # [unix]
    - USE_NNPACK # [unix]
    - USE_QNNPACK # [unix]
    - BUILD_TEST # [unix]
    - USE_PYTORCH_METAL_EXPORT # [osx]
    - USE_COREML_DELEGATE # [osx]
    - _GLIBCXX_USE_CXX11_ABI # [unix]

test:
 imports:
    - torch

about:
  home: http://pytorch.org/
  license: BSD 3-Clause
  license_family: BSD
  license_file: LICENSE
  summary: PyTorch is an optimized tensor library for deep learning using GPUs and CPUs.
```
</details>

Which means that numpy is required at build time, but not runtime. Therefore, we should move the line you added to the `run_constraint`.

hmaarrfk: I'm not an expert with llvm. I can't comment on that.

hmaarrfk:
gah sorry https://github.com/conda-forge/conda-forge-ci-setup-feedstock/pull/231

Tobias-Fischer:
Awesome, thanks for that! Seems like you got the osx-* build all to pass which is great :)

recipe/meta.yaml
```diff
@@ -81,12 +80,13 @@ outputs:
         - requests
         - future
         - six
-        - cffi
         - mkl-devel {{ mkl }}   # [x86]
         - libcblas * *_mkl      # [x86]
         - libcblas              # [not x86]
         - liblapack             # [not x86]
         - openblas              # [not x86]
+        - libgomp   # [linux]
```
hmaarrfk, line 88: i was only going to ask about this line with respect to the CUDA builds.

hmaarrfk: ok lets skip the builds this time, and maybe upload them over time later

hmaarrfk: ninja was already removed as a dependency so i htink the main issue is resolved via a repo-data patch

hmaarrfk (approved):

recipe/meta.yaml
hmaarrfk: Congrats on the move!

Tobias-Fischer:
Can confirm this PR fixed https://github.com/conda-forge/norse-feedstock/pull/11

hmaarrfk:
great!

Issue #151, opened 2022-12-05
Pytorch 2.0


h-vetinari:
Pytorch 2.0 recently got [announced](https://pytorch.org/get-started/pytorch-2.0/), and it looks like it'll bring lots of goodies.

However, with the strong focus on the new compiler backend, things are predictably getting heavier for the build. Seems the default backend will be triton, which we have in conda-forge [already](https://github.com/conda-forge/triton-feedstock), but so far only for linux-x64 + CUDA (no CPU version). That feedstock is also pushing hard against the 6h limit.

Release is planned March 2023.

If we have the capacity, I think it'd make sense to start playing with this, perhaps even pushing into a dev branch, so that we might have a chance to provide feedback on critical issues before release.

hmaarrfk:
We should likely create a branch for 1.13 and migrate it. While publishing rc to a dev channel. We can do this, even if just for the CPU builds.

h-vetinari:
IMO 1.13 can stay on main for a while, we only need a dev branch and `pytorch_dev` label. OTOH I think we could already start before the rc phase

hmaarrfk:
We could make a release candidate branch branch and release to a non main channel if you want

h-vetinari:
We can start with a PR; if we have the resources to create builds and push them to a dev channel, all the better

h-vetinari:
https://github.com/pytorch/pytorch/releases/tag/v2.0.0 was released today! :)

Tobias-Fischer:
Hi all,
Happy to help out if you need a hand. @hmaarrfk: You have quite a few draft/WIP PRs. Will it make sense to get some merged first, as this would simplify the version update? Or is it best to base a v2 branch on top of main?

hmaarrfk:
I've mostly given up on my experimental PRs.  main is the best to follow.

Tobias-Fischer:
Just letting you know that I’ve been working on this and I’m pretty close, at least for the non-cuda Linux-64 build. Will create a PR next week

h-vetinari:
Closed by https://github.com/conda-forge/pytorch-cpu-feedstock/pull/165

h-vetinari:
Triton discussion continues in https://github.com/conda-forge/pytorch-cpu-feedstock/issues/166

PR #165, opened 2023-04-02
Bump to PyTorch 2.0.0

Tobias-Fischer:
Checklist


Tobias-Fischer:
Fixes #151

Tobias-Fischer:
Still struggling with some issues.

1) For the non-cuda build on linux-64, at least locally I was running into issues where the run dependencies of the pytorch package are not pulled in by the pytorch-cpu package (which is really weird) and leads for the build to fail as `sympy` cannot be found when `import torch`.

2) Locally the cuda build fails with
```
conda.CondaMultiError: The package for cudnn located at /home/conda/feedstock_root/build_artifacts/pkg_cache/cudnn-8.4.1.50-hed8a83a_0
appears to be corrupted. The path 'bin/.cudnn-post-link.sh'
specified in the package manifest cannot be found.

The package for cudnn located at /home/conda/feedstock_root/build_artifacts/pkg_cache/cudnn-8.4.1.50-hed8a83a_0
appears to be corrupted. The path 'include/cudnn.h'
specified in the package manifest cannot be found.
```

3) Have not tested the osx builds.


Let's see what happens in CI.

Tobias-Fischer:
It seems like this is working quite well :)

Unfortunately linux-aarch64 errors with:
```
ImportError: /home/conda/feedstock_root/build_artifacts/pytorch-recipe_1680469918528/_test_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_p/lib/python3.11/site-packages/torch/lib/libtorch_cpu.so: undefined symbol: _ZNK6google8protobuf7Message11GetTypeNameEv
```

Any ideas where this issue might be from @conda-forge/pytorch-cpu @conda-forge/libprotobuf @conda-forge/protobuf? It seems like it could be related to the gcc version (https://github.com/NVIDIA-AI-IOT/torch2trt/issues/53)? I can only find this issue on conda-forge but it does not seem to apply here (mixing defaults and conda-forge channels): https://github.com/conda-forge/paraview-feedstock/issues/23

Several issues hint to towards issues with onnx/caffe2.

hmaarrfk:
my guess is that somebody is re-exporting the symbols publicly. It may be the vendored onnx?

hmaarrfk:
Maybe related to: https://github.com/pytorch/pytorch/blob/5d62d1255778b53ece16c79fd842cd42eca31f93/CMakeLists.txt#L175

Tobias-Fischer:
Looks like this is now ready for review. Could someone else please test a cuda build locally?

Tobias-Fischer:
Thanks @hmaarrfk!

We can test the gpu builds with
```
>>> import torch

>>> torch.cuda.is_available()
True

>>> torch.cuda.device_count()
1
```

hmaarrfk:
```
>>> import torch
>>> torch.cuda.is_available()
True
>>> torch.cuda.device_count()
1
>>> print(torch.__version__)
2.0.0.post200
```

is the `post` expected?

hmaarrfk:
Builds starting to upload https://anaconda.org/mark.harfouche/pytorch/files if people want a more thorough test.

Tobias-Fischer:
There are a few instances of `post200` when searching on the PyTorch repo: https://github.com/search?q=repo%3Apytorch%2Fpytorch+post200&type=issues

Not sure what it refers to

hmaarrfk:
Of course, pytorch needs to have a custom version numbering system ^_^

https://github.com/pytorch/pytorch/blob/73b06a0268bb89c09a86f16fa0f72818baa4b250/tools/generate_torch_version.py#L51

The use the CONDA_BUILD_NUMBER and add it to the version number.

I can't fully trace it, but it seems to be related to the build number. 

Lets maybe flag this as an issue, but I don't want to patch too much at this stage.

Tobias-Fischer:
I don't think it's a big deal, is it? If you really wanted to, we could change https://github.com/Tobias-Fischer/pytorch-cpu-feedstock/blob/5f00d8033d7cc23eb9831c74dc038fe3e4047562/recipe/build_pytorch.sh#LL56 to 0 instead.

hmaarrfk:
I think it is ok as is.
Ah thnk you. i searched pytorch's repo for that statement but forgot that maybe we just set it...

Eventually, i think somebody's version check will be broken but we can deal with it later.

Tobias-Fischer:
So far no one has complained, and we've had the same situation in past conda-forge builds:
```
>>> torch.__version__
'1.13.0.post200'
```

hmaarrfk:
Yeah. I agree. I think it's fine.

Just waiting for the builds at this stage.

h-vetinari:
I thought that 2.0 needed Triton as a backend, or is that optional...?

hmaarrfk:
hm, sorry i forgot to test for triton, is there a "test" you would like be to try on a GPU?

hmaarrfk:
linux compiling just takes "time" but it is easy to start, so if somebody wants to add triton support, i can rebuild linux.

Tobias-Fischer:
- I am a bit confused what is depending on what. As far as I can see, pytorch depends on torchtriton, which in turn seems to depend on pytorch. 
- Is there any difference between torchtriton and triton?
- We already package an old version of triton in conda-forge and have a PR open for version 2.0.0 (https://github.com/conda-forge/triton-feedstock/pull/2). Would this 2.0.0 version be suitable as dependency for here?

Tobias-Fischer:
Another issue seems to be that I can't install a recent version of torchvision and pytorch=2 side-by-side; I manage to kill mamba ;)

```
> mamba create -n pytorch21 pytorch=2 torchvision=0.14
python: /home/conda/feedstock_root/build_artifacts/mamba-split_1680002410624/work/libmamba/src/core/satisfiability_error.cpp:1767: mamba::{anonymous}::TreeExplainer::node_t mamba::{anonymous}::TreeExplainer::concat_nodes(const std::vector<long unsigned int>&): Assertion `std::all_of( ids.begin(), ids.end(), [&](auto id) { return m_pbs.graph().node(ids.front()).index() == m_pbs.graph().node(id).index(); } )' failed.
```

If I don't specify the torchvision version, it pulls in a very old torchvision which is definitely not compatible ..

hmaarrfk:
pytorch run exports itself, so torchvision has to be rebuilt.

PR #168, opened 2023-04-20
Generic blas

ngam:
Checklist


recipe/meta.yaml
```diff
@@ -42,6 +42,8 @@ outputs:
         - python                                 # [build_platform != target_platform]
         - cross-python_{{ target_platform }}     # [build_platform != target_platform]
         - numpy                                  # [build_platform != target_platform]
+        - libcblas              # [not x86 and build_platform != target_platform]
+        - liblapack             # [not x86 and build_platform != target_platform]
```
h-vetinari: Unnecessary because we're always natively compiling x86 (i.e. build_platform _is_ the same as target_platform).

```suggestion
        - libcblas                               # [build_platform != target_platform]
        - liblapack                              # [build_platform != target_platform]
```

isuruf: Please keep `not x86`. Dependencies changing whether cross or native is wrong.

h-vetinari: OK... I'm a bit surprised, the "repeat necessary host-deps in build for cross-compilation with `# [build_platform != target_platform]`" pattern is used for almost all our cross-compiled packages.

isuruf: Wait, this is in `build`? Why do we need `libblas` in `build`?

h-vetinari: Yes, it's in build. I haven't checked why it's been added or whether it's needed, but it does happen occasionally that some build-dep needs to discover something at build-time (even though we try to point them to host wherever possible).

PR #203, opened 2023-11-07
Update 2.1.0 again

hmaarrfk:
@jakirkham the same error happened again.

The CI should show it too:

```
  -- Found Threads: TRUE
  CMake Error at cmake/public/cuda.cmake:65 (message):
    Found two conflicting CUDA installs:

    V12.0.76 in
    '/home/conda/feedstock_root/build_artifacts/pytorch-recipe_1699320420794/_h_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placeh/targets/x86_64-linux/include'
    and

    V12.0.76 in
    '/home/conda/feedstock_root/build_artifacts/pytorch-recipe_1699320420794/_build_env/targets/x86_64-linux/include'
  Call Stack (most recent call first):
    cmake/Dependencies.cmake:44 (include)
    CMakeLists.txt:722 (include)
```
Checklist


jakirkham:
Thanks Mark! 🙏

It appears this is due to a change introduced in [PyTorch 2.1.0's own CTK detection logic]( https://github.com/pytorch/pytorch/pull/82695/files#diff-8e5cb190cc46be808993381a31fe9c027705d356b6bc0460368c0310ae82b273R61-R66 ). Unfortunately this check is problematic

In the Conda case, we implement a splayed layout. This means build tools (like those listed in `requirements/build`) live in one path and libraries that are linked to (like those listed in `requirements/host`) live in another path. This is common when supporting cross-compilation as we do with Conda. However this means there are cases where we do need to use some things from both paths for different reasons

It looks like @ax3l ran into this issue in the HPC SDK use case and proposed a fix ( https://github.com/pytorch/pytorch/pull/108932 ). Idk if that will work for us

In terms of our needs here, maybe just patching the check out altogether would be reasonable to get the build working

Perhaps this would be a good opportunity to discuss with @peterbell10 whether we can come up with a better check in PyTorch that works for splayed layout use cases

hmaarrfk:
ok lets try again, i took a brute force approach because we can do a bit of meta building

recipe/meta.yaml
```diff
@@ -25,6 +25,11 @@ source:
   git_tag: v{{ version }}
   patches:
     - nvml.patch  # [cuda_compiler_version == "11.2"]
+    # The patch below is probably OK for other versions too, but it is untested with them
+    # https://github.com/conda-forge/pytorch-cpu-feedstock/pull/203#issuecomment-1797352452
+    # https://github.com/pytorch/pytorch/pull/82695/files#diff-8e5cb190cc46be808993381a31fe9c027705d356b6bc0460368c0310ae82b273R61-R66
+    # https://github.com/pytorch/pytorch/pull/108932
+    - 0001-Allow-splayed-layouts.patch  # [cuda_compiler_version == "12.0"]
```
jakirkham, line 32: Think we might need this

```suggestion
    - patches/0001-Allow-splayed-layouts.patch  # [cuda_compiler_version == "12.0"]
```

hmaarrfk: thanks . i got it locally. i cleaned up a little more.

hmaarrfk:
It now fails with:
```
  CMake Error at cmake/public/cuda.cmake:64 (message):
    Failed to find nvToolsExt
  Call Stack (most recent call first):
    cmake/Dependencies.cmake:44 (include)
    CMakeLists.txt:722 (include)


```

recipe/meta.yaml
jakirkham: Thanks Mark! 🙏

jakirkham:
Thanks Mark! 🙏

Think that refers to [this check
added in the same PyTorch PR]( https://github.com/pytorch/pytorch/pull/82695/files#diff-8e5cb190cc46be808993381a31fe9c027705d356b6bc0460368c0310ae82b273R68-R70 )

As noted under [the `CUDA::nvToolsExt` doc in CMake]( https://cmake.org/cmake/help/latest/module/FindCUDAToolkit.html#cuda-toolkit-nvtoolsext ), this is a deprecated target by CMake (and NVIDIA) as it comes from NVTX 2, NVTX 3 has superseded it

Am guessing PyTorch doesn't use NVTX 2 (as this check was new in that PR). Meaning that this was purely a build configuration check

So think we can remove those lines as well

hmaarrfk:
what about this:
https://github.com/pytorch/pytorch/pull/82695/files#diff-8e5cb190cc46be808993381a31fe9c027705d356b6bc0460368c0310ae82b273R218

hmaarrfk:
can i remove it too?

recipe/build_pytorch.sh
```diff
@@ -128,20 +128,28 @@ if [[ ${cuda_compiler_version} != "None" ]]; then
...
+        export CUDA_TOOLKIT_ROOT_DIR=$CUDA_HOME
     elif [[ ${cuda_compiler_version} == 11.2 ]]; then
         export TORCH_CUDA_ARCH_LIST="3.5;5.0;6.0;6.1;7.0;7.5;8.0;8.6+PTX"
+        export CUDA_TOOLKIT_ROOT_DIR=$CUDA_HOME
     elif [[ ${cuda_compiler_version} == 11.8 ]]; then
         export TORCH_CUDA_ARCH_LIST="3.5;5.0;6.0;6.1;7.0;7.5;8.0;8.6;8.9+PTX"
+        export CUDA_TOOLKIT_ROOT_DIR=$CUDA_HOME
     elif [[ ${cuda_compiler_version} == 12.0 ]]; then
         export TORCH_CUDA_ARCH_LIST="5.0;6.0;6.1;7.0;7.5;8.0;8.6;8.9;9.0+PTX"
+        # CUDA_TOOLKIT_ROOT_DIR no longer needed for 12.0
```
jakirkham, line 152:
```suggestion
        # $CUDA_HOME not set in CUDA 12.0. Using $PREFIX
        export CUDA_TOOLKIT_ROOT_DIR="${PREFIX}"
```

jakirkham (commented):
Hmm...think you make a good point

In looking through the code see [some references to NVTX 2]( https://github.com/pytorch/pytorch/blob/ddfe5725342b0c0f707222879ca9dac305f97210/torch/csrc/profiler/stubs/cuda.cpp#L3 )

We could override [where PyTorch looks for NVTX]( https://github.com/pytorch/pytorch/blob/ddfe5725342b0c0f707222879ca9dac305f97210/cmake/Modules/FindCUDAToolkit.cmake#L1051 ), but maybe it is best to point PyTorch to where the bulk of the CTK content is in the native compilation case

It's possible this will need tweaking further in the build, but hopefully this unblocks configuration

hmaarrfk:
ok well i pushed my changes, feel free to push anything if you can. going to slee....

jakirkham:
Thanks Mark! 🙏

Have a good night

jakirkham:
It looks like Peter added a fix upstream ( https://github.com/pytorch/pytorch/pull/113174 ). Thanks Peter! 🙏

Maybe we can give that a try

hmaarrfk:
it doesn't address the nvtools issue

Tobias-Fischer:
There are two upsteam PRs that solve the nvtools issue; I am not sure which one is preferable. See https://github.com/pytorch/pytorch/issues/101135 and PRs https://github.com/pytorch/pytorch/pull/97582 and https://github.com/pytorch/pytorch/pull/106763

recipe/build_pytorch.sh
```diff
@@ -128,20 +128,29 @@ if [[ ${cuda_compiler_version} != "None" ]]; then
...
     elif [[ ${cuda_compiler_version} == 11.2 ]]; then
         export TORCH_CUDA_ARCH_LIST="3.5;5.0;6.0;6.1;7.0;7.5;8.0;8.6+PTX"
+        export CUDA_TOOLKIT_ROOT_DIR=$CUDA_HOME
     elif [[ ${cuda_compiler_version} == 11.8 ]]; then
         export TORCH_CUDA_ARCH_LIST="3.5;5.0;6.0;6.1;7.0;7.5;8.0;8.6;8.9+PTX"
+        export CUDA_TOOLKIT_ROOT_DIR=$CUDA_HOME
     elif [[ ${cuda_compiler_version} == 12.0 ]]; then
         export TORCH_CUDA_ARCH_LIST="5.0;6.0;6.1;7.0;7.5;8.0;8.6;8.9;9.0+PTX"
+        # $CUDA_HOME not set in CUDA 12.0. Using $PREFIX
+        export CUDA_TOOLKIT_ROOT_DIR="${PREFIX}"
```
jakirkham, line 153: Maybe we need to point PyTorch to the particular target directory. Usually this hasn't been needed with CMake builds, but there is at least one case where we had to go this route ( https://github.com/conda-forge/kaldi-feedstock/pull/39 )

Here's [the layout logic]( https://github.com/conda-forge/cuda-nvcc-feedstock/blob/394f3817decd9316269b5a40e5076b8bcc94dc33/recipe/activate.sh#L5-L7 ) for context

```suggestion
        if [[ "${target_platform}" == "linux-64" ]]; then
            export CUDA_TOOLKIT_ROOT_DIR="${PREFIX}/targets/x86_64-linux"
        elif [[ "${target_platform}" == "linux-ppc64le" ]]; then
            export CUDA_TOOLKIT_ROOT_DIR="${PREFIX}/targets/ppc64le-linux"
        elif [[ "${target_platform}" == "linux-aarch64" ]]; then
            export CUDA_TOOLKIT_ROOT_DIR="${PREFIX}/targets/sbsa-linux"
        else
            echo "Unsupported architecture by CUDA"
            exit 1
        fi
```

jakirkham: Also if we would rather just do the `linux-64` logic, that's fine. Just wanted to spell out the other architectures for completeness

hmaarrfk:
ok builds are incoming:

My test was to use 
```python
import torch
a = torch.randn(1024 * 1024 * 1024, device='cuda')
a + 1
```

and watch the memory on my cuda device grow using nvtop

recipe/meta.yaml
```diff
@@ -33,6 +44,7 @@ build:
   skip: true  # [cuda_compiler_version == "10.2"]
   skip: true  # [cuda_compiler_version == "11.0"]
   skip: true  # [cuda_compiler_version == "11.1"]
+  skip: true  # [cuda_compiler_version != "12.0"]
```
jakirkham, line 47: Guessing this is a temporary measure to reign in builds offline. Is that right?

jakirkham:
Nice work Mark! 🥳

@peterbell10 do the patches here look upstreamable to you? Or are there similar approaches that upstream could take that would alleviate the need for these?

recipe/meta.yaml
hmaarrfk: yes. I can add a note about it. but I'm not thrilled about using 8OSX CPUs for 8-12 hours for this.

Tobias-Fischer: Do you mean the CI CPUs? If so, can't we stop the Azure builds manually? Or do you need help with local builds?

hmaarrfk: I don't know how to stop the build manually. I just feel that somebody will remember to enable the osx builds.....

Tobias-Fischer: Agreed :)

Tobias-Fischer (approved):

hmaarrfk:
[log_files.zip](https://github.com/conda-forge/pytorch-cpu-feedstock/files/13315316/log_files.zip)

jakirkham:
Were Linux ARM packages built as well?

Issue #202, opened 2023-11-06
Mega builds instead of tiny builds


hmaarrfk:
### Comment:

In https://github.com/conda-forge/pytorch-cpu-feedstock/issues/108 a proposal was made to split up the package into tiny chunks.

One challenge I faced is that this really goes against upstream's versionning practices for certain things like XNNPACK. 

The alternative I see is to actually go the other way, and to go for "megabuilds".

1. Build libtorch in a split package.
2. Do not delete the output files
3. Build the different python versions

This will tip us over the edge on the 6 hour, but we are already there. The current problem is rebuilding the CUDA stuff, is done 5 times, for python 3.8 through 3.12..... not fun....

just an idea I wanted to throw out there in case somebody had time to try it out.

RaulPPelaez:
I belive this broke some of my feedstocks that depend on torch. I started getting this error on new releases:
```shell
["  ERROR (torchmd-net,lib/python3.11/site-packages/torchmdnet/extensions/torchmdnet_extensions.so): .. but ['conda-forge/linux-64::libtorch==2.1.0=cuda120_h86db2e7_302'] not in reqs/run, (i.e. it is overlinking) (likely) or a missing dependency (less likely)", "  ERROR (torchmd-net,lib/python3.11/site-packages/torchmdnet/extensions/torchmdnet_extensions.so): .. but ['conda-forge/linux-64::libtorch==2.1.0=cuda120_h86db2e7_302'] not in reqs/run, (i.e. it is overlinking) (likely) or a missing dependency (less likely)", "  ERROR (torchmd-net,lib/python3.11/site-packages/torchmdnet/extensions/torchmdnet_extensions.so): .. but ['conda-forge/linux-64::libtorch==2.1.0=cuda120_h86db2e7_302'] not in reqs/run, (i.e. it is overlinking) (likely) or a missing dependency (less likely)"]
```
https://github.com/conda-forge/torchmd-net-feedstock/pull/11

Stumbled on to this when I read about the previously non-existent libtorch package. I had to add libtorch as a dependency.

njzjz:
I tried to migrate from pytorch to libtorch at https://github.com/conda-forge/abacus-feedstock/pull/25, since it's a C++-only project. Everything was good, but I had to manually pin libtorch, since there is no global pin.

hmaarrfk:
xref: https://github.com/conda-forge/conda-forge-pinning-feedstock/pull/5412

PR #210, opened 2023-12-01
Build libtorch once and pytorch multiple times

isuruf:
Implements https://github.com/conda-forge/pytorch-cpu-feedstock/issues/202

Also adds self-hosted runners


hmaarrfk:
does this work? even without the self hosted runnings i'm interested in merging this.

isuruf:
> does this work? even without the self hosted runnings i'm interested in merging this.

Yes, but it sends macOS builds over the limit. 5 hours for libtorch + 0.5 hours * 5 for python buids.

hmaarrfk:
> 5 hours for libtorch 

i notice that it spends ALOT of time copying over headers. 

Do you think it might be worth doing minimal patching to the cmakefiles to copy over the headers with libtorch?

recipe/meta.yaml
```diff
@@ -40,22 +40,138 @@ source:
 
 build:
   number: {{ build }}
+  string: cuda{{ cuda_compiler_version | replace('.', '') }}_h{{ PKG_HASH }}_{{ PKG_BUILDNUM }}  # [cuda_compiler_version != "None"]
+  string: cpu_{{ blas_impl }}_h{{ PKG_HASH }}_{{ PKG_BUILDNUM }}                                 # [cuda_compiler_version == "None"]
+  detect_binary_files_with_prefix: false
+  run_exports:
+    - {{ pin_subpackage('libtorch', max_pin='x.x') }}
+  ignore_run_exports_from:
+    - python *                               # [not osx]
```
hmaarrfk, line 49: maybe it would be better to define a new variable like: megabuild

and condition on that? that way we could keep a comment as to why osx is not a megabuild.

hmaarrfk:
Any idea for osx+Intel?

isuruf:
This is ready now. The linux failure is because of the large logs and we have to rebuild only that manually.

hmaarrfk:
> of the large logs

can we use a "sed filter" to just cleanup some of the log?

isuruf:
> can we use a "sed filter" to just cleanup some of the log?

Not sure. I'll keep working on it.

isuruf:
@hmaarrfk, let's merge this and keep building 11.2 manually for now?

hmaarrfk:
yes that is ok.

but is there a material need for users to have this new build for 11.2?

or can we just merge and then rebuild 11.2 when we need to?

isuruf:
We can just merge and then rebuild 11.2 when we need to

hmaarrfk:
are you just able to merge + cancel these jobs and save the cycles?

we can keep the build number bumped, but I also feel bad just copying GBs of data on anaconda's servers for nothing.

hmaarrfk:
if you can, please merge at will

isuruf:
> we can keep the build number bumped, but I also feel bad just copying GBs of data on anaconda's servers for nothing.

I mean it's good to have the libtorch package for users which reduces the load on anaconda servers and users.

hmaarrfk:
sorry for not reviewing this so thoroughly, does this new package place the libraries in the correct location, or does it still put them under some `lib/python3.X` location?

isuruf:
It puts them in `$PREFIX/lib` in `libtorch` and then there are symlinks in `lib/python3.x/site-packages/pytorch/lib` to `$PREFIX/lib` in `pytorch`.

hmaarrfk:
> It puts them in $PREFIX/lib in libtorch and then there are symlinks in lib/python3.x/site-packages/pytorch/lib to $PREFIX/lib in pytorch.

great!

Did you find the symlinks necessary? or is it out of an abundance of caution

isuruf:
It is necessary. For eg: https://github.com/pytorch/pytorch/blob/main/torch/__init__.py#L168-L176
We could patch those, but symlinks are easier.

hmaarrfk:
hmm fun. ok thanks!

very thorough as usual.

my computers are kinda free, but my brain power isn't.

I can try to rebuild for 11.2 at some point, but I just dont have the capacity to think about this week for such an old version of cuda.

hmaarrfk:
merge at will!

h-vetinari:
This is awesome! 🚀🥳🍾

hmaarrfk:
log files for cuda 11.2: [log_files.zip](https://github.com/conda-forge/pytorch-cpu-feedstock/files/13964970/log_files.zip)

isuruf:
Looks like uploading from GHA failed.

hmaarrfk:
Do you want me to start building the Linux ones locally?

isuruf:
It was a token issue. Fixed the token issue and restarted the jobs. Packages should be up now.

hmaarrfk:
great!

PR #238, opened 2024-05-16
 Rebuild for numpy 2.0; use larger runners on GPU server

h-vetinari:
2.3.0 should be compatible with numpy 2.0 already; for some reason, the bot stumbles over the templated outputs here (probably an old failure before stdlib was added), so let's do it by hand.


h-vetinari:
> @h-vetinari Can you link to the log error on the bot? I want to investigate to see why it cannot process this recipe.

This is what was on the status page:
```
bot error (
[bot CI job](https://github.com/regro/cf-scripts/actions/runs/8962449813)
): main: Traceback (most recent call last):
  File "/home/runner/work/cf-scripts/cf-scripts/cf-scripts/conda_forge_tick/auto_tick.py", line 1271, in _run_migrator
    migrator_uid, pr_json = run(
                            ^^^^
  File "/home/runner/work/cf-scripts/cf-scripts/cf-scripts/conda_forge_tick/auto_tick.py", line 237, in run
    migrator.run_pre_piggyback_migrations(recipe_dir, feedstock_ctx.attrs, **kwargs)
  File "/home/runner/work/cf-scripts/cf-scripts/cf-scripts/conda_forge_tick/migrators/core.py", line 268, in run_pre_piggyback_migrations
    mini_migrator.migrate(recipe_dir, attrs, **kwargs)
  File "/home/runner/work/cf-scripts/cf-scripts/cf-scripts/conda_forge_tick/migrators/cstdlib.py", line 206, in migrate
    sections = _slice_into_output_sections(lines, attrs)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/runner/work/cf-scripts/cf-scripts/cf-scripts/conda_forge_tick/migrators/libboost.py", line 48, in _slice_into_output_sections
    raise RuntimeError("Could not find all output sections in meta.yaml!")
RuntimeError: Could not find all output sections in meta.yaml!
```

h-vetinari:
In the CPU megabuild, we built libtorch and several pytorches, but then fail on the py38 build with:
```
  CMake Error in torch/CMakeLists.txt:
    Imported target "numpy::numpy" includes non-existent path

      "/home/conda/feedstock_root/build_artifacts/libtorch_1715855634503/_h_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_pl/lib/python3.8/site-packages/numpy/_core/include"

    in its INTERFACE_INCLUDE_DIRECTORIES.  Possible reasons include:

    * The path was deleted, renamed, or moved to another location.

    * An install or uninstall procedure did not complete successfully.

    * The installation package was faulty and references files it does not
    provide.
```

Also, we haven't caught a 4xl runner in over 10h, so I'll try if 2xl works better...

beckermr:
Ahhh that error message is old. When I put in the PR to improve that section of the bot, I changed the error message. See the code here: https://github.com/regro/cf-scripts/blob/master/conda_forge_tick/migrators/libboost.py#L107.

h-vetinari:
> Ahhh that error message is old.

That's why I had said in the OP "probably an old failure before stdlib was added", because once that string is found in the recipe, the piggyback isn't even attempted. :)

beckermr:
Ahhhh thanks @h-vetinari. I didn't grok your comment fully! sorry!

h-vetinari:
Thanks for the help @isuruf! 

Sidenote @beckermr: I don't understand why the two jobs with the green checkmarks aren't skipped completely (as in: smithy shouldn't generate a job for them at all; all outputs are skipped for them AFAICT).

recipe/build_pytorch.sh
```diff
@@ -1,6 +1,7 @@
 set -x
 if [[ "$megabuild" == true ]]; then
-  source $RECIPE_DIR/build.sh
+  source $RECIPE_DIR/build_libtorch.sh
+  mkdir -p $SP_DIR/torch
```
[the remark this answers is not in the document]
h-vetinari: The mkdir bit? That's literally what failed for https://github.com/conda-forge/pytorch-cpu-feedstock/pull/238/commits/b4d9ffb2a44c7e1dfad9dc3cf69693b4238145af:
```
$SRC_DIR/build_pytorch.sh: line 7: pushd: $PREFIX/lib/python3.8/site-packages/torch: No such file or directory
```
Also `mkdir -p` is a no-op if the folder exists already, so I don't see the problem?

The build-script rename is a consequence of https://github.com/conda-forge/pytorch-cpu-feedstock/pull/238/commits/981a04d378bda35a2f2852cd0a448f0bcc9d56e6 and I prefer to keep it.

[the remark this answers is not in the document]
isuruf: If that file is not there, something else failed. With the line you added, the error is only apparent at the testing phase instead of build phase.

h-vetinari:
> Which ci support files?

The two with the green check marks... because they get skipped immediately. IOW:
* linux_64_blas_implgenericc_compiler_version11cuda_compilernvcccuda_compiler_version11.8cxx_compiler_version11
* linux_64_blas_implgenericc_compiler_version12cuda_compilercuda-nvcccuda_compiler_version12.0cxx_compiler_version12

These should both be skipped by:
```
skip: true  # [cuda_compiler_version != "None" and linux64 and blas_impl != "mkl"]
```
which is applied for all outputs/stages.

isuruf:
> Also, we haven't caught a 4xl runner in over 10h, so I'll try if 2xl works better...

You need to request access for those. See https://github.com/conda-forge/admin-requests/blob/main/examples/example-open-gpu-server.yml

recipe/conda_build_config.yaml
```diff
@@ -1,18 +1,16 @@
...
 
 blas_impl:
-  - mkl                        # [x86 or x86_64]
+  - mkl                         # [x86_64]
   - generic
 
 github_actions_labels:
-- cirun-openstack-gpu-large    # [x86 or x86_64]
+- cirun-openstack-gpu-2xlarge   # [x86_64]
 - cirun-openstack-cpu-large
```
h-vetinari, line 14: One of the big issues here with the cirun sizes is that we need to distinguish CUDA builds from non-CUDA builds, but that's really hard to do in a local pinning, which is I guess why the distinction on `x86_64` through `blas_impl` is kind of a work-around. But it also means we cannot set the right CI runner for aarch+CUDA.

I've tried to come up with a solution for this, but it would need changes to the global pinning: https://github.com/conda-forge/conda-forge-pinning-feedstock/pull/5903

.ci_support/linux_64_blas_implmklc_compiler_version12cuda_compilerNonecuda_compiler_versionNonecxx_compiler_version12.yaml
```diff
@@ -27,7 +27,7 @@ cxx_compiler_version:
 docker_image:
 - quay.io/condaforge/linux-anvil-cos7-x86_64
 github_actions_labels:
-- cirun-openstack-gpu-2xlarge
+- cirun-openstack-cpu-large
```
h-vetinari, line 30: This was wrong before (CPU build getting a GPU runner)

.ci_support/linux_aarch64_c_compiler_version12cuda_compilercuda-nvcccuda_compiler_version12.0cxx_compiler_version12.yaml
```diff
@@ -31,7 +31,7 @@ cxx_compiler_version:
 docker_image:
 - quay.io/condaforge/linux-anvil-cos7-x86_64
 github_actions_labels:
-- cirun-openstack-cpu-large
+- cirun-openstack-gpu-2xlarge
```
h-vetinari, line 34: Same here, aarch+CUDA build getting the smaller CPU runner

h-vetinari (commented):
Main changes that illustrate/motivate https://github.com/conda-forge/pytorch-cpu-feedstock/pull/238/commits/7e0cc8e607d178c45b5a91b51bd998569f6fb280 & https://github.com/conda-forge/conda-forge-pinning-feedstock/pull/5903

(unfortunately, I couldn't get the superfluous generic+CUDA variants to disappear)

recipe/conda_build_config.yaml
```diff
@@ -1,24 +1,56 @@
...
+  - 11.8                    # [linux]
+  - 12.0                    # [linux]
+docker_image:                                   # [linux]
+  - quay.io/condaforge/linux-anvil-cos7-x86_64  # [linux]
+  - quay.io/condaforge/linux-anvil-cuda:11.8    # [linux]
+  - quay.io/condaforge/linux-anvil-cos7-x86_64  # [linux]
+github_actions_labels:                          # [linux]
+  - cirun-openstack-cpu-large                   # [linux]
+  - cirun-openstack-gpu-2xlarge                 # [linux]
+  - cirun-openstack-gpu-2xlarge                 # [linux]
```
isuruf, line 56: Let's just use `cirun-openstack-cpu-2xlarge` everywhere to keep this file manageable.

recipe/meta.yaml
```diff
@@ -36,128 +43,142 @@ source:
...
-  detect_binary_files_with_prefix: false
-  run_exports:
-    - {{ pin_subpackage('libtorch', max_pin='x.x') }}
-  ignore_run_exports_from:
-    - python *                               # [megabuild]
-    - numpy *                                # [megabuild]
-    - cross-python_{{ target_platform }}     # [megabuild and build_platform != target_platform]
-  ignore_run_exports:
-    - python *                               # [megabuild]
-    - numpy *                                # [megabuild]
```
isuruf, line 50: Why all these changes?

h-vetinari: It's just an indentation change. If you switch off whitespace visibility, you'll see what I mean.

recipe/conda_build_config.yaml
h-vetinari: OK, fine by me.

recipe/meta.yaml
h-vetinari: See in particular: https://github.com/conda-forge/pytorch-cpu-feedstock/pull/238/commits/981a04d378bda35a2f2852cd0a448f0bcc9d56e6?w=1

isuruf: No, it's not only an indentation change. You are changing the top level recipe from `libtorch` to `pytorch-recipe` and then moving build scripts to an output. We need a fan-out build here and it's only a coincidence that the recipe in this PR works. Please revert.

h-vetinari: Fan-outs also work between outputs, a lot of other recipes rely on that (e.g. arrow).

I'm really against reusing the same name for both package & output. It's really non-obvious what's happening, and it's much better to create a separate output.

h-vetinari:
Hm, this is still failing to get the 2xl runners due to:
```
{
    "error": "User not authorized for the requested runners, user not present in 'users' list."
}
```
I don't see how https://github.com/Quansight/open-gpu-server/blob/main/access/conda-forge-users.json would distinguish user rights between different runners. Are we the first ones to try something with 2xl runners?

recipe/meta.yaml
isuruf: I'm against relying on an undocumented feature where there's an documented feature to be used.

h-vetinari: Can you point me to some documentation for which fan-out builds are supported (or not)?

It's been working like that for years and has several users that I'm aware of. I'd certainly raise an issue in conda-build as a regression if any of this breaks. It's a natural application of the variant algebra from my POV. One key (`python:`) with several entries is used for one output but not the other, so build that one first and then loop through the python versions for the dependent one.

In contrast, the way that reusing the package name for an output actually works is all but obvious for anyone who's not already familiar with this (very rare) pattern - e.g. where are the tests (not defined before this PR, but if they were, they'd come before the output they're testing). A lot of people will look at the pytorch feedstock. It's more important to me that the recipe be understandable than how conda-build does it behind the scenes.

isuruf: Can you please stop arguing over every little thing? You are making conda-forge less fun for me and to at least one other person I know of. I'm going to stop commenting here. I'll leave it upto @conda-forge/pytorch-cpu to merge this. Please add yourself as a maintainer here as I will not contribute to this feedstock anymore.

h-vetinari: > Can you please stop arguing over every little thing?

Excuse me? You opened this discussion and insisted on your view. Am I not allowed to defend mine?

> You are making conda-forge less fun for me and to at least one other person I know of.

I'm sorry to hear that. As you well know, I've reached out to you repeatedly over the last couple of years to find a way to work together productively, asking what you'd like me (not) to do, and you've given me no indication besides (paraphrasing) "do your homework better". While I've tried that to the absolute limit my free time allows, it's frankly been challenging to work with your curt-verging-on-combative responses towards me across many a feedstock...

> Please add yourself as a maintainer here as I will not contribute to this feedstock anymore.

... but yet I keep trying, because you are an absolutely essential contributor/reviewer. So I'll continue trying to do my best to find ways to work together harmoniously. Still, I don't see why we cannot resolve a difference of opinion through a technical discussion (rather than throwing out the baby with the bathwater), or why your opinion (on a feedstock where we're both not officially maintainers) should win by default over mine.

h-vetinari:
Alright, this is back to fully green CI, no manual builds required! 🥳 

PTAL @conda-forge/pytorch-cpu :)

recipe/build_libtorch.sh
```diff
@@ -84,7 +84,8 @@ if [[ "$CONDA_BUILD_CROSS_COMPILATION" == 1 ]]; then
     export COMPILER_WORKS_EXITCODE__TRYRUN_OUTPUT=""
 fi
 
-export MAX_JOBS=${CPU_COUNT}
+# reduce parallelism to avoid getting OOM-killed
+export MAX_JOBS=4
```
hmaarrfk, line 88: would such a change be OK? It is useful to be able to able to run faster if you can.
```suggestion
if [[ "${CI}" == "github_actions" ]]; then
# h-vetinari/hmaarrfk -- May 2024
# reduce parallelism to avoid getting OOM-killed
# This has been tuned for our 
# runner: cirun-openstack-gpu-2xlarge
# that has 32GB of RAM and 8 CPU cores
  export MAX_JOBS=4
else
  export MAX_JOBS=${CPU_COUNT}
fi
```

i'm not sure if we can switch based on `${github_actions_labels}` but if we can that would be better

recipe/build_libtorch.sh
```diff
@@ -97,6 +98,11 @@ fi
 if [[ "$PKG_NAME" == "pytorch" ]]; then
   PIP_ACTION=install
   sed "s/3.12/$PY_VER/g" build/CMakeCache.txt.orig > build/CMakeCache.txt
+  if [[ "$PY_VER" == "3.8" ]]; then
```
hmaarrfk, line 101:
```suggestion
  # h-vetinari - May 2024 
  # We use a fan-out build to avoid the long rebuild of libtorch
  # However, the location of the numpy headers changes between python 3.8
  # and 3.9+ since numpy 2.0 only exists for 3.9+
  if [[ "$PY_VER" == "3.8" ]]; then
```

recipe/meta.yaml
hmaarrfk: @isuruf I ask that you please reconsider your statement:

> as I will not contribute to this feedstock anymore.

it has been great to work with you on this project (and many others).

I think that review comments are often difficult to contextualize from people sitting on the other side of a screen, often on the other side of world. I know that even short sentences like:

> Why all these changes?

and 

> It's just an indentation change [...]

were well thought out by the both of you. However, after a long day of work, I know that both of these might seem like you are brushing off the comment of the other.

I may be "injecting" my own thoughts in here, but my reading of your interactions make it seem like the fuller sentences would have been:

> Interesting approach, I'm surprised that this isn't causing a full rebuild of `libtorch` many times. It seems that the fan-out build is still being triggered even though it isn't at the top level. I don't think this is well documented and may disappear in a future version of conda-build. Is it possible to the old syntax? If so I would prefer that approach.

and

> I've been using fan-outs in many other recipes such as grpc, if this is a quirk in the conda-build, I would rather get it resolved. May you point me to the documentation you are referring to so that I may appropriately ask for clarification in a centralized location?


While it may be hard to reconcile, I know you are both going above an beyond with your personal time to review these recipes and keep conda-forge as a whole in good working order.


----------

Personally, I've avoided these so called "fan-out" builds and historically have been moving away from naming multi-output recipes the same name as one of the outputs due to the rapid pace of their development and the fact that they seem to change feature set every so often (in my perspective)


 I tried to search through both:
* https://docs.conda.io/projects/conda-build/en/stable/
* https://conda-forge.org/docs/

for the word "fan-out" and I couldn't find anything of substance. It would be helpful for me. I agree with Isuru's thoughts that we should use "documented" features (especially for these nearly impossible to build recipes), and I feel like a comment pointing to the appropriate literature might be appropriate at the top level of the recipe.


As a way to move forward could:

* h-vetinari is it possible to move to the old syntax for now on this recipe?
* isuruf are you able to point us to the documentation on fan-out builds so that we can refer to it here and ask questions in the most appropriate repo?

recipe/meta.yaml
```diff
@@ -36,128 +43,142 @@ source:
...
+        - pytorch-cpu ==99999999       # [cuda_compiler_version != "None"]
+        - pytorch {{ version }} cuda{{ cuda_compiler_version | replace('.', '') }}_*_{{ PKG_BUILDNUM }}  # [cuda_compiler_version != "None"]
+        - pytorch {{ version }} cpu_{{ blas_impl }}_*_{{ PKG_BUILDNUM }}                                 # [cuda_compiler_version == "None"]
+
+    test:
+      commands:
+        # libraries
+        {% for each_lib in libs %}
+        - test -f $PREFIX/lib/{{ each_lib }}${SHLIB_EXT}
+        {% endfor %}
```
hmaarrfk, line 180:
```suggestion
        # libraries
        {% for each_lib in [ 'libc10', 'libshm', 'libtorch', 'libtorch_cpu', 'libtorch_global_deps'] %}
        - echo testing for the presence $PREFIX/lib/{{ each_lib }}${SHLIB_EXT}   # [unix]
        - test -f $PREFIX/lib/{{ each_lib }}${SHLIB_EXT}   # [unix]
        {% endfor %}
        {% for each_lib in ['libc10_cuda', 'libcaffe2_nvrtc', 'libtorch_cuda', 'libtorch_cuda_linalg'] %}
        - echo testing for the presence of $PREFIX/lib/{{ each_lib }}${SHLIB_EXT}  # [unix and cuda_compiler_version != "None"]
        - test -f $PREFIX/lib/{{ each_lib }}${SHLIB_EXT}  # [unix and cuda_compiler_version != "None"]
        {% endfor %}
```

In this case, I don't think the added indirection of having the libraries listed at the top "saves" any cognitive load.

It would be good to list of libraries being tested next to the test itself.
The echo command has also helped me troubleshoot recipes since the failed `test -f` doesn't often print out what file was missing.

hmaarrfk (commented):
This is really amazing work, and I'm impressed you overcame some of the tricky python 3.8/numpy2 compatibility challenges. 

I'm hoping that we can:

1. Add a few comments for the tricks you used.
2. Enable "manual builds" to still run. Its really useful to test things without a CI locally on a more powerful machine before testing.
3. Cleanup some nits of the syntax in the `meta.yaml` (I'm trying not to trigger builds on the large CI resources)
4. Revert the fan-out syntax changes while we get to the bottom of the documented/undocumented features of this feature in conda-build.

recipe/meta.yaml
```diff
@@ -15,8 +15,15 @@
 {% set cuda_major = environ.get("cuda_compiler_version", "11.8").split(".")[0] | int %}
 {% endif %}
 
+{% set libs = [
+    'libc10', 'libshm', 'libtorch', 'libtorch_cpu', 'libtorch_global_deps'
+] + (cuda_compiler_version != "None") * [
+    'libc10_cuda', 'libcaffe2_nvrtc', 'libtorch_cuda', 'libtorch_cuda_linalg'
+] %}
+
+
```
hmaarrfk, line 24:
```suggestion
```

Hopefully these github comments make it easy for you to address my nits.

recipe/meta.yaml
h-vetinari: Thanks for the mediation effort @hmaarrfk! I can go back to the previous way for now, and refer to documentation, if any exists (I'm not aware of one, so I consider both ways equally undocumented, but if that's what's necessary to move on, so be it).

Otherwise we can open an issue in conda-build for clarification.

recipe/build_libtorch.sh
h-vetinari: > i'm not sure if we can switch based on `${github_actions_labels}` but if we can that would be better

That value would be populated according to CBC also in a local build, so that wouldn't achieve your goal AFAIU.

h-vetinari:
Just noticing the following
```
SafetyError: The package for libtorch located at /home/conda/feedstock_root/build_artifacts/pkg_cache/libtorch-2.3.0-cuda118_h9e56e6c_301
appears to be corrupted. The path 'bin/torch_shm_manager'
has an incorrect size.
  reported size: 43368 bytes
  actual size: 49545 bytes
Verifying transaction: ...working... done

SafetyError: The package for libtorch located at /home/conda/feedstock_root/build_artifacts/pkg_cache/libtorch-2.3.0-cuda118_h9e56e6c_301
appears to be corrupted. The path 'lib/libc10.so'
has an incorrect size.
  reported size: 1094504 bytes
  actual size: 1158313 bytes

SafetyError: The package for libtorch located at /home/conda/feedstock_root/build_artifacts/pkg_cache/libtorch-2.3.0-cuda118_h9e56e6c_301
appears to be corrupted. The path 'lib/libc10_cuda.so'
has an incorrect size.
  reported size: 614840 bytes
  actual size: 646881 bytes

SafetyError: The package for libtorch located at /home/conda/feedstock_root/build_artifacts/pkg_cache/libtorch-2.3.0-cuda118_h9e56e6c_301
appears to be corrupted. The path 'lib/libcaffe2_nvrtc.so'
has an incorrect size.
  reported size: 21088 bytes
  actual size: 26289 bytes

SafetyError: The package for libtorch located at /home/conda/feedstock_root/build_artifacts/pkg_cache/libtorch-2.3.0-cuda118_h9e56e6c_301
appears to be corrupted. The path 'lib/libshm.so'
has an incorrect size.
  reported size: 48192 bytes
  actual size: 56665 bytes

SafetyError: The package for libtorch located at /home/conda/feedstock_root/build_artifacts/pkg_cache/libtorch-2.3.0-cuda118_h9e56e6c_301
appears to be corrupted. The path 'lib/libtorch.so'
has an incorrect size.
  reported size: 14816 bytes
  actual size: 17129 bytes

SafetyError: The package for libtorch located at /home/conda/feedstock_root/build_artifacts/pkg_cache/libtorch-2.3.0-cuda118_h9e56e6c_301
appears to be corrupted. The path 'lib/libtorch_cpu.so'
has an incorrect size.
  reported size: 210126256 bytes
  actual size: 214128009 bytes

SafetyError: The package for libtorch located at /home/conda/feedstock_root/build_artifacts/pkg_cache/libtorch-2.3.0-cuda118_h9e56e6c_301
appears to be corrupted. The path 'lib/libtorch_cuda.so'
has an incorrect size.
  reported size: 1343656712 bytes
  actual size: 1353944689 bytes

SafetyError: The package for libtorch located at /home/conda/feedstock_root/build_artifacts/pkg_cache/libtorch-2.3.0-cuda118_h9e56e6c_301
appears to be corrupted. The path 'lib/libtorch_cuda_linalg.so'
has an incorrect size.
  reported size: 845552 bytes
  actual size: 901641 bytes

SafetyError: The package for libtorch located at /home/conda/feedstock_root/build_artifacts/pkg_cache/libtorch-2.3.0-cuda118_h9e56e6c_301
appears to be corrupted. The path 'lib/libtorch_global_deps.so'
has an incorrect size.
  reported size: 14984 bytes
  actual size: 17201 bytes
```

I think we might need to clean some build caches in the megabuild.

recipe/meta.yaml
h-vetinari: > The echo command has also helped me troubleshoot recipes since the failed `test -f` doesn't often print out what file was missing.

I've never seen that happen... The failing line (with all the substitutions done) gets printed in the logs. I've left out the echos for now, we can add them if we do ever run into a problem with that.

For reference, here's the log from the last run, i.e. without any echos:
```
export PREFIX=/home/conda/feedstock_root/build_artifacts/pytorch-recipe_1716077967897/_test_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_p
export SRC_DIR=/home/conda/feedstock_root/build_artifacts/pytorch-recipe_1716077967897/test_tmp
+ test -f $PREFIX/lib/libc10.so
+ test -f $PREFIX/lib/libshm.so
+ test -f $PREFIX/lib/libtorch.so
+ test -f $PREFIX/lib/libtorch_cpu.so
+ test -f $PREFIX/lib/libtorch_global_deps.so
+ test -f $PREFIX/lib/libc10_cuda.so
+ test -f $PREFIX/lib/libcaffe2_nvrtc.so
+ test -f $PREFIX/lib/libtorch_cuda.so
+ test -f $PREFIX/lib/libtorch_cuda_linalg.so
+ exit 0
```

h-vetinari:
Luckily, the SafetyErrors are gone now too. Anything else you'd want here @hmaarrfk? 🙃

hmaarrfk:
Thanks @h-vetinari, this looks great! I think it addresses most parties' concerns. 

We can consider the build refactor in a separate PR.

@isuruf so this comment doesn't get lost, do take a chance to consider it if you ever happen to back to this PR.
https://github.com/conda-forge/pytorch-cpu-feedstock/pull/238#discussion_r1606051235

PR #231, opened 2024-04-05
feat: yet another attempt to add windows builds

baszalmstra:
Checklist
* [x] Used a [personal fork of the feedstock to propose changes](https://conda-forge.org/docs/maintainer/updating_pkgs.html#forking-and-pull-requests)
* [x] Bumped the build number (if the version is unchanged)
* [x] Reset the build number to `0` (if the version changed)
* [x] [Re-rendered]( https://conda-forge.org/docs/maintainer/updating_pkgs.html#rerendering-feedstocks ) with the latest `conda-smithy` (Use the phrase <code>@<space/>conda-forge-admin, please rerender</code> in a comment in this PR for automated rerendering)
* [ ] Ensured the license file is being packaged.

Fixes #32 

This PR is another attempt to add Windows builds (see #134) . 

For now I disabled all other builds to be able to test the windows part first. I made this PR draft so we don't accidentally merge it.

Issue #229, opened 2024-04-04
State of pytorch


baszalmstra:
### Comment:

Hey dear maintainers and contributors.

Conda is used quite a bit in the ML ecosystem. It's a great option because installing pytorch for your particular system should be just `conda install pytorch` which would install a pytorch version targeting your version of Cuda, ROCm, or CPU architecture. I love this!

However, there are some issues that I have been facing.

- There are no packages for Windows (#32)
- The feedstocks latest version `2.1.2` is (almost) 4 months old. (#221)
- No ROCm support (#198)

I see people switching to pip or using the pytorch channel which introduce their own problems.

It looks like a number of these issues are related to infrastructure problems. I would love to contribute to improve this but Im not entirely sure where to start so Im opening this issue to start a conversation and get in contact with the people who do.

hmaarrfk:
which one of the challenges would you like to tackle. In my experience it is going to be nearly impossible to tackle them all at once. 

Focus on one area of "need" and work toward it.

If you can get the builds to run on Azure, that is really easy to main, we click merge.

I might have dragged my feet on a merge request:
https://github.com/conda-forge/pytorch-cpu-feedstock/pull/225#issuecomment-2011052510

I know i'm being "selfish" but co-instability of pytorch and tensorflow is important to me. so i felt like between the two:

1. Most updated versions
2. co-installable tensorflow + pytorch

i chose `#2`. I could be swayed (I have my own channels that I maintain for this reason) but addressing the "tensorflow problem" is also related to this 
https://github.com/conda-forge/tensorflow-feedstock/issues/378

baszalmstra:
Personally, I think not having windows builds is a big reason to not use pytorch from conda-forge. Even if the version is not completely up to date, not having a version available is worse. ;)

However, I have been building pytorch (-cpu) with rattler-build on windows and it requires a lot of resources, I think that is also the reason the effort in https://github.com/conda-forge/pytorch-cpu-feedstock/pull/134 was halted?

But I would be happy to revive that PR if things have changed in the mean time?

hmaarrfk:
> Even if the version is not completely up to date, not having a version available is worse. ;)

I'm not sure that is true. Our migration infrastructure requires all packages to be up to date for all platforms.

So if somebody contributes a windows package one day, then gets pulled away due to other priorities, then pytorch is effectively halted.

rattler likely isn't the cause of the slowdown. 

The usage of "git" to clone the large repo is slow.
[...]

Issue #155, opened 2023-01-12
Feedstocks that pin pytorch-cpu


h-vetinari:
We should clean these up, because they're preventing users from picking CUDA-enabled pytorch for no good reason:

* [ ] [eeyore](https://github.com/conda-forge/eeyore-feedstock/blob/main/recipe/meta.yaml#L25)
* [ ] [htm-pytorch](https://github.com/conda-forge/htm-pytorch-feedstock/blob/main/recipe/meta.yaml#25)
* [ ] [nlpaug](https://github.com/conda-forge/nlpaug-feedstock/blob/main/recipe/meta.yaml#L27)
* [ ] [sagemaker-pytorch-serving-container](https://github.com/conda-forge/sagemaker-pytorch-serving-container-feedstock/blob/main/recipe/meta.yaml#L30)
* [ ] [sparktorch](https://github.com/conda-forge/sparktorch-feedstock/blob/main/recipe/meta.yaml#23)
* [ ] [torchbiggraph](https://github.com/conda-forge/torchbiggraph-feedstock/blob/main/recipe/meta.yaml#L35)
* [ ] [yews](https://github.com/conda-forge/yews-feedstock/blob/main/recipe/meta.yaml#L28)

Found by this [search](https://github.com/search?q=org%3Aconda-forge+-+pytorch-cpu&type=code)

h-vetinari:
Thanks @mgorny for tackling the remaining feedstocks here! 🙏

hmaarrfk:
all checked off!!! Time to close?

h-vetinari:
> Found by this [search](https://github.com/search?q=org%3Aconda-forge+%2F-+pytorch-%5Bcg%5Dpu%2F+path%3Arecipe%2F*.yaml&type=code)

New ones keep appearing 😑

hmaarrfk:
many are test time dependencies to try to stop (large) gpu packages from getting pulled in. we are ok with that right?

h-vetinari:
Yeah, test dependencies are ok of course. It's only about avoiding to force a choice of GPU or non-GPU on ambient packages, so stuff in `run:` basically

hmaarrfk:
i can't find any from the search. is your point that you often go back to the search and check up on this issue?

h-vetinari:
"often" is not the word I'd use. Maybe once or twice a year? Last time was in February, which yielded one new issue:
* https://github.com/conda-forge/isofit-feedstock/issues/56

PR #247, opened 2024-07-12
remove `pytorch-{cpu,gpu}` compat outputs

h-vetinari:
While [reviewing](https://github.com/conda-forge/pytorch-cpu-feedstock/pull/231#discussion_r1675188481) another PR, I checked the situation with regarding `pytorch-{cpu,gpu}` again, and IMO it's way past time to remove them. Quoting from there

>  They were purely for compatibility with the old naming on the pytorch channel, but we've lost track of them completely.
> 
> In the pytorch channel, `pytorch-cpu` hasn't had a new [build](https://anaconda.org/pytorch/pytorch/files) in 5(!) years, and the gpu variant doesn't exist anymore(?!). There's [`pytorch-cuda`](https://anaconda.org/pytorch/pytorch-cuda/files), but that's only a meta-package for the various CUDA components, not pytorch itself. You can check out the upstream [recipe](https://github.com/pytorch/builder/blob/main/conda/pytorch-nightly/meta.yaml), which now also features a build variant.

We might want to wait for a new minor version to do this.

PS. We've been emitting linter hints for these since https://github.com/conda-forge/conda-forge-pinning-feedstock/commit/c00c43a2f809d3836e83cbf721bd93d58366924f

### DIFF
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -1,5 +1,5 @@
 {% set version = "2.3.1" %}
-{% set build = 1 %}
+{% set build = 2 %}
 
 {% if cuda_compiler_version != "None" %}
 {% set build = build + 200 %}
@@ -142,16 +142,11 @@ requirements:
     # GPU requirements without run_exports
     - {{ pin_compatible('cudnn') }}                       # [cuda_compiler_version != "None"]
   run_constrained:
-    # These constraints ensure conflict between pytorch and
-    # pytorch-cpu 1.1 which we built before conda-forge had GPU infrastructure
-    # built into place.
-    # https://github.com/conda-forge/pytorch-cpu-feedstock/issues/65
-    - pytorch-cpu =={{ version }}  # [cuda_compiler_version == "None"]
-    - pytorch-gpu ==99999999       # [cuda_compiler_version == "None"]
-    - pytorch-gpu =={{ version }}  # [cuda_compiler_version != "None"]
-    - pytorch-cpu ==99999999       # [cuda_compiler_version != "None"]
     - pytorch {{ version }} cuda{{ cuda_compiler_version | replace('.', '') }}_*_{{ PKG_BUILDNUM }}  # [cuda_compiler_version != "None"]
     - pytorch {{ version }} cpu_{{ blas_impl }}_*_{{ PKG_BUILDNUM }}                                 # [cuda_compiler_version == "None"]
+    # obsolete compat outputs that have been removed
+    - pytorch-cpu <0.0a0
+    - pytorch-gpu <0.0a0
 
 # these tests are for the libtorch output below, but due to
 # a particularity of conda-build, that output is defined in
@@ -277,14 +272,9 @@ outputs:
         - __cuda  # [cuda_compiler_version != "None"]
         - libtorch {{ version }}
       run_constrained:
-        # These constraints ensure conflict between pytorch and
-        # pytorch-cpu 1.1 which we built before conda-forge had GPU infrastructure
-        # built into place.
-        # https://github.com/conda-forge/pytorch-cpu-feedstock/issues/65
-        - pytorch-cpu =={{ version }}  # [cuda_compiler_version == "None"]
-        - pytorch-gpu ==99999999       # [cuda_compiler_version == "None"]
-        - pytorch-gpu =={{ version }}  # [cuda_compiler_version != "None"]
-        - pytorch-cpu ==99999999       # [cuda_compiler_version != "None"]
+        # obsolete compat outputs that have been removed
+        - pytorch-cpu <0.0a0
+        - pytorch-gpu <0.0a0
 
     test:
       requires:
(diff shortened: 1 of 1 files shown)


h-vetinari:
(CI stopped to conserve resources; we can discuss first)

hmaarrfk (approved):
But i would also just want to cancel on main.

Triggering a whole new 500MB download for end users that don't need it is silly.

hmaarrfk:
Lets wait for a few more stakeholders.

h-vetinari:
It would also be good to take another stab at fixing https://github.com/conda-forge/pytorch-cpu-feedstock/issues/155; I updated the list and opened some issues for hopefully higher visibility.

It's worth noting though that we already added linter hints for `pytorch-{cpu,gpu}` as of https://github.com/conda-forge/conda-forge-pinning-feedstock/commit/c00c43a2f809d3836e83cbf721bd93d58366924f almost a year ago

isuruf:
How do you propose end users select the CPU/GPU version now?

hmaarrfk:
> How do you propose end users select the CPU/GPU version now?

```
mamba install pytorch
```

should allow them to get the "best version" pretty consistently. 

otherwise

```
mamba install pytorch=*=cuda*
# vs
mamba install pytorch=*=cpu_*
```

isuruf:
```
mamba install pytorch=*=cuda*
# vs
mamba install pytorch=*=cpu_*
```

This is a bit verbose than
```
mamba install pytorch-gpu
# vs
mamba install pytorch-cpu
```
What's the harm in continuing to support these convenience outputs? It's not a burden to maintain these.


pytorch channel has the convenience outputs
```
mamba install pytorch pytorch-cuda=x
vs
mamba install pytorch cpuonly
```

h-vetinari:
The original intention was to match the naming of the pytorch channel so users don't accidentally coinstall the same package with different names from different channels.

So now the upstream naming has diverged, I think those outputs are barely worth it anymore.

Already, the ideal should be IMO that user choice is unnecessary, in the sense that they get the "best" implementation for their system; for the cases where a choice is necessary, the build string option is not amazing, but possible (similar to how we're advising users to set `libblas=*=*<impl>`)

If we want to continue providing convenience wrappers for that, then IMO we should match the pytorch naming (except `cpuonly`), e.g. there's also `pytorch-mutex` that they use for that. 

Matching `pytorch-cuda` would be attractive in principle, if only it didn't involve a completely different version scheme...

Long term, I'd wish for a better [API](https://github.com/conda/conda/issues/11053) to express these things, without messing with pointless outputs or build strings.

isuruf:
IMO, just dropping this output and wasting user's time to figure out that you now need to say `pytorch=*=*_cpu` is not worth it
especially since having these are not a huge maintenance burden.

hmaarrfk (changes requested):
I'm removing my approval since isuruf's comments reminded me of an old issue I created that still has no resolution

https://github.com/conda-forge/conda-forge.github.io/issues/1894

to my knowledge, today, there is no way to specify:

````
pytroch-gpu 2.3.1
```
should just be rewritten as
```
pytorch 2.3.1=cuda*
```
when the users types
```
mamba update --all
```

In our application, we try to identify these snags, and give users command like

```
# Do not replace the first conda command with mamba
conda uninstall --force --yes --offline old_package_name
mamba install "new_package_name>=some_lower_bound"
mamba update --all --yes
```

While I am ok with people in new environments being asked to use the new names of packages, the fact that we can't "update" old environments really bothers me since it leaves users in very strange "stuck states" of the solver.

rgommers:
I also think `pytorch-cpu`/`pytorch-gpu` is significantly nicer than using build strings. Removing them will be a breaking change for some users (that'd include me; xref https://github.com/conda-forge/pytorch-cpu-feedstock/issues/229#issuecomment-2493511045 for an example). And with the `pytorch` channel being deprecated now, we should just do whatever seems best for Conda-forge. 

I suggest closing this PR and keeping things as is.

h-vetinari:
Alright, let's close this then. My POV is still

> Long term, I'd wish for a better [API](https://github.com/conda/conda/issues/11053) to express these things, without messing with pointless outputs or build strings.

rgommers:
> Long term, I'd wish for a better [API](https://github.com/conda/conda/issues/11053) to express these things, without messing with pointless outputs or build strings.

No argument there from me - a nicer new syntax would be great, and if we get that then we can revisit this PR I think.

function2-llx:
Hello, I've been watching your discussion on removing both `pytorch-cpu` and `pytorch-gpu` in favor of `pytorch`, I appreciate your effort and would like to provide a suggestion.

Now that the official PyTorch Anaconda channel is deprecated, and users are guided to the conda-forge alternative in the [official announcement](https://github.com/pytorch/pytorch/issues/138506). However, they are explicitly suggesting `pytorch-cpu` and `pytorch-gpu` in the guidance. I'm afraid this will make more users migrate to these packages that are supposed to be deprecated some days, and significantly burden your efforts on finally removing them. In addition, there will probably be more items added to the list in https://github.com/conda-forge/pytorch-cpu-feedstock/issues/155.

Therefore, I strongly suggest that you communicate with the PyTorch team and make sure their announcement / documentation / tutorial are adjusted accordingly.

h-vetinari:
Thanks for the comment - it's OK though, we've decided to keep the convenience outputs anyway.

Issue #254, opened 2024-08-28
PyTorch 2.4.0 Package Not Installable w/ CUDA 12 on Python 3.12 Linux x86_64


iamthebot:
### Solution to issue cannot be found in the documentation.

- [X] I checked the documentation.

### Issue

On a Linux x86_64 machine:
```
CONDA_OVERRIDE_CUDA=12 conda install pytorch
...
Channels:
 - conda-forge
Platform: linux-64
Collecting package metadata (repodata.json): done
Solving environment: done

## Package Plan ##

  environment location: /home/alfredo_luque/.airconda-environments/devel--demo--alfredo_luque--airconda_tutorial--v0.0.1

  added / updated specs:
    - pytorch


The following packages will be downloaded:

    package                    |            build
    ---------------------------|-----------------
    fsspec-2024.6.1            |     pyhff2d567_0         130 KB  conda-forge
    numpy-2.1.0                |  py312h1103770_0         8.0 MB  conda-forge
    ------------------------------------------------------------
                                           Total:         8.1 MB

The following NEW packages will be INSTALLED:

  _sysroot_linux-64~ conda-forge/noarch::_sysroot_linux-64_curr_repodata_hack-3-h69a702a_16
  cuda-version       conda-forge/noarch::cuda-version-11.8-h70ddcb2_3
  cudatoolkit        conda-forge/linux-64::cudatoolkit-11.8.0-h4ba93d1_13
  cudnn              conda-forge/linux-64::cudnn-8.9.7.29-hbc23b4c_3
  filelock           conda-forge/noarch::filelock-3.15.4-pyhd8ed1ab_0
  fsspec             conda-forge/noarch::fsspec-2024.6.1-pyhff2d567_0
  gmp                conda-forge/linux-64::gmp-6.3.0-hac33072_2
  gmpy2              conda-forge/linux-64::gmpy2-2.1.5-py312h1d5cde6_1
  icu                conda-forge/linux-64::icu-75.1-he02047a_0
  jinja2             conda-forge/noarch::jinja2-3.1.4-pyhd8ed1ab_0
  kernel-headers_li~ conda-forge/noarch::kernel-headers_linux-64-3.10.0-h4a8ded7_16
  libabseil          conda-forge/linux-64::libabseil-20240116.2-cxx17_he02047a_1
  libblas            conda-forge/linux-64::libblas-3.9.0-23_linux64_openblas
[...]

PR #255, opened 2024-08-29
Enable linux ppc64le

jeongseok-meta:
Checklist

### DIFF
--- a/conda-forge.yml
+++ b/conda-forge.yml
@@ -26,5 +26,6 @@ github_actions:
 provider:
   linux_64: github_actions
   linux_aarch64: azure
+  linux_ppc64le: azure
   win_64: github_actions
 test: native_and_emulated
--- a/conda-forge.yml
+++ b/conda-forge.yml
@@ -10,6 +10,7 @@ bot:
   - v2.5.x
 build_platform:
   linux_aarch64: linux_64
+  linux_ppc64le: linux_64
   osx_arm64: osx_64
 conda_build:
   pkg_format: '2'
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -138,7 +138,7 @@ requirements:
     - cusparselt
     - libcublas-dev
     - libcudss-dev
-    - libcufile-dev  # [linux]
+    - libcufile-dev  # [linux and not ppc64le]
     - libcufft-dev
     - libcurand-dev
     - libcusolver-dev
@@ -289,7 +289,7 @@ outputs:
         - cusparselt
         - libcublas-dev
         - libcudss-dev
-        - libcufile-dev  # [linux]
+        - libcufile-dev  # [linux and not ppc64le]
         - libcufft-dev
         - libcurand-dev
         - libcusolver-dev
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -1,6 +1,6 @@
 # if you wish to build release candidate number X, append the version string with ".rcX"
 {% set version = "2.6.0" %}
-{% set build = 2 %}
+{% set build = 3 %}
 
 # Use a higher build number for the CUDA variant, to ensure that it's
 # preferred by conda's solver, and it's preferentially
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -122,7 +122,7 @@ requirements:
     - rsync     # [unix]
   host:
     # GPU requirements
-    - cudnn                           # [cuda_compiler_version != "None"]
+    - cudnn                           # [cuda_compiler_version != "None" and not ppc64le]
     - nccl                            # [cuda_compiler_version != "None" and linux]
     - magma                           # [cuda_compiler_version != "None"]
     - cuda-version {{ cuda_compiler_version }}  # [cuda_compiler_version != "None"]
@@ -135,9 +135,9 @@ requirements:
     - cuda-nvtx-dev
     - cuda-nvml-dev
     - cuda-profiler-api
-    - cusparselt
+    - cusparselt     # [not ppc64le]
     - libcublas-dev
-    - libcudss-dev
+    - libcudss-dev   # [not ppc64le]
     - libcufile-dev  # [linux and not ppc64le]
     - libcufft-dev
     - libcurand-dev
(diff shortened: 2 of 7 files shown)


conda-forge.yml
```diff
@@ -25,6 +25,7 @@ os_version:
 provider:
   linux_64: github_actions
   linux_aarch64: azure
+  linux_ppc64le: default
```
jakirkham, line 28: Think this can be `azure`

Though please make sure to add `linux_ppc64le` to `build_platform` earlier in this file

jakirkham:
Thanks Jeongseok! 🙏 

This looks like a good set of changes

For the remaining CI jobs, think you will need to accept the ToS, which is pretty similar to other CIs ToS. This amounts to making a PR like this one adding your username to the list like so: https://github.com/Quansight/open-gpu-server/pull/37

Once that is done we can see how CI progresses

jeongseok-meta:
Done! https://github.com/jeongseok-meta/open-gpu-server/pull/1

jeongseok-meta:
Done! https://github.com/Quansight/open-gpu-server/pull/39

jakirkham:
Great, thank you! 🙏

Asked Jaime to take a look as he typically reviews those 🙂

Am wondering if it is worth grabbing this workaround for NumPy builds in PyTorch: https://github.com/conda-forge/pytorch-cpu-feedstock/issues/254#issuecomment-2319195557

The reason being we discovered ARM builds likely need that workaround. So think if that is true, it is probably also true for Power

What do you think?

Should add it is ok if you prefer to wait and see how CI goes first. We do now cover that in the test suite after having seen it before. So we will know either way if it is an issue

https://github.com/conda-forge/pytorch-cpu-feedstock/blob/6dd85b3f85a72371b2d3ccf6a386de67e61d667e/recipe/meta.yaml#L332

hmaarrfk:
I suggest adding back the early failure I am suggesting in:
https://github.com/conda-forge/pytorch-cpu-feedstock/pull/256/commits/292cd8ee3f16bb337ec74b73a213d3ea39a22efe

you might be faster at debuging what is going on, but that early test might be more than adequate

jakirkham:
Agree that sounds like a good plan. Thanks for putting it together Mark! 🙏

Commented over there 🙂

jeongseok-meta:

jeongseok-meta:
@conda-forge-admin, please restart ci

hmaarrfk:
please add:

```
skip: true  # [py!=310]
```

to the top level build section

if you are not ready yet to get the GPU enabled, please add an other line with:

```
skip: true   # [cuda_compiler_version!=None]
```

and rerender

hmaarrfk:
@conda-forge-admin please rerender

hmaarrfk:
sorry about that, lets just skip everything except for ppc64le for now and then re-optimize the builds.

jeongseok-meta:
Sure, feel free to do anything that makes sense to you or take over this PR. Thank you for helping!

hmaarrfk:
no i'm struggling with aarch64 already. i won't be able to take over, but this feedstock just hogs the CIs, then you end up building linux64 for 6 hours instead of running ppc64le which you want!

hmaarrfk:
maybe try rerendering locally, rerendering with cuda is always slow...

hmaarrfk:
@conda-forge-admin please rerender

One more rerender. I am allowed to the GPU runners, but you can simply use azure until the 6 hour timeout is hit.

hmaarrfk:
@conda-forge-admin please rerender 


sigh... always one more thing to rerender

hmaarrfk:
alright, you have the CIs now. happy tuning!

hmaarrfk:
i'm also assuming that you have access to a powerful linux machine with docker.

To build locally

```
python build-locally.py
```

and it should walk you through things.

hmaarrfk:
While the builds have started, I have a feeling you are hitting the same problems I am with aarch 
image

See:
https://github.com/conda-forge/pytorch-cpu-feedstock/pull/256

Depending on what you are trying to prove at this stage (numpy detection vs others), you might want to pull in my patch that causes this to fail loudly.

recipe/meta.yaml
```diff
@@ -57,6 +57,10 @@ build:
   # 2024/04 This build seems to fail due to disk space on the CIs
   # We can potentially re-enable it at a later time
   skip: true  # [cuda_compiler_version == "11.8" and aarch64]
+  skip: true  # [py!=310]
+  skip: true  # [osx]
+  skip: true  # [aarch64]
+  skip: true  # [linux64]
```
hmaarrfk, line 63:
```suggestion
  skip: true  # [linux64]
  skip: true  # [cuda_compiler_version != "None"]
```

While you are hitting failures in the CPU build, it would be nice to just cancel the cuda builds too ;)

jakirkham:
Happy New Year! 🥳 

Noticed some conflicts crept in due to recent merges

To fix conflicts, would recommend...

1. Check this PR branch locally
2. Attempt merge
3. Resolve conflicts on directory `recipe/` only (`git mergetool recipe` works well)
4. `git rm -f '.ci_support/linux_*' '.ci_support/osx_*' '.ci_support/win_*'`
5. [Re-render]( https://conda-forge.org/docs/maintainer/updating_pkgs/#rerendering-feedstocks )
6. Finish merge and push changes

jeongseok-meta:

jeongseok-meta:

jeongseok-meta:
Thanks for the work and discussion here. I re-evaluated this against current `main`, and I think this PR is no longer the right path for ppc64le support.

This branch was based on the PyTorch 2.6 / CUDA 12.4 era, while the feedstock is now on PyTorch 2.10 with CUDA 12.9/13.0 variants. The PR is draft, conflicting, and current CUDA ppc64le package support does not appear to cover the CUDA stack used by the feedstock today.

If ppc64le support is revisited, I think it should be restarted as a fresh CPU-only draft from current `main`, with CUDA explicitly skipped on ppc64le, generated files rerendered separately, and full build evidence before merge discussion.

Closing this to avoid carrying a stale direction. Thanks again for helping move the investigation forward.

Issue #273, opened 2024-10-08
Help-Wanted Priority List


hmaarrfk:
By no means a strict order but here is my opinion on where we can use the most help. Jump in where you can.

While feature additions are helpful, we prioritize maintainability and compatibility with the rest of the conda-forge ecosystem.

Please try to discuss implementation details in the dedicated issue or PR.

### Maintenance
* "RC Builds": current version is 2.5.0rcX

### Maintainability
* Consider Upstream's split build: https://github.com/conda-forge/pytorch-cpu-feedstock/issues/272
* Unvendoring: https://github.com/conda-forge/pytorch-cpu-feedstock/issues/108 
* cuDSS: https://github.com/conda-forge/pytorch-cpu-feedstock/issues/269

### User support
* https://github.com/conda-forge/pytorch-cpu-feedstock/pull/255
* https://github.com/conda-forge/pytorch-cpu-feedstock/issues/275

### Feature addition
* Triton support: https://github.com/conda-forge/pytorch-cpu-feedstock/issues/166
* MKLDNN + CUDA: https://github.com/conda-forge/pytorch-cpu-feedstock/issues/233
* Other blas + CUDA: https://github.com/conda-forge/pytorch-cpu-feedstock/issues/177
* GDS: https://github.com/conda-forge/pytorch-cpu-feedstock/issues/257
* https://github.com/conda-forge/pytorch-cpu-feedstock/issues/269
* Optimized builds CUDA 12.2 or 12.4 https://github.com/conda-forge/pytorch-cpu-feedstock/issues/270
* Identifying the broken builds OSX arm + Pytorch https://github.com/conda-forge/pytorch-cpu-feedstock/issues/260

## Windows Stream

Mostly this needs a dedicated champion for the Windows cause. The large builds make this incompatible with the conda-forge philosophy of building in the open using automated bots.

* A powerful windows CI for >6 hr builds -- https://github.com/conda-forge/pytorch-cpu-feedstock/issues/268
    * May be substituted with a benevolent contributor.
    * Contributor must be responsive and adhere to [CFEP03](https://github.com/conda-forge/cfep/blob/main/cfep-03.md)
* Windows CPU + system ABSL
    * A proof of concept Windows build was demonstrated in 2024 in https://github.com/conda-forge/pytorch-cpu-feedstock/pull/231 We just prefer to use conda-forge ABSL instead of rebuilding it for each feedstock.
[...]

PR #280, opened 2024-10-28
Allow builds of rc-tagged commits

danpetry:
Checklist

- Addresses "rc builds" point [here](https://github.com/conda-forge/pytorch-cpu-feedstock/issues/273)
- A step towards aligning [Anaconda's](https://github.com/AnacondaRecipes/pytorch-feedstock) and conda-forge's recipe
- This has been used in the above recipe for a while; suggest disabling the CI, merging, and verifying next time it's needed to conserve resources, but I'm happy with a different approach if desired.


danpetry:

beckermr (changes requested):
Any rc builds need to go to non-main label.

hmaarrfk:
> Any rc builds need to go to non-main label.

can we do this with some kind of "jinja" syntax? i don't think so right?

beckermr:
I think it might be able to zip in the channel config in the conda build config.

danpetry:
Are you thinking something like this?
```
channel_targets:
  - conda-forge main
  - conda-forge rc
rc:
  - none
  - 5
zip_keys:
  -
    - channel_targets
    - rc
```
I don't think you'd want to build a release candidate and a normal release in parallel, though. I can't see a way to make channel_targets conditional, since it needs to be in the cbc.yaml.

What are your requirements around this in general? We use it to prepare for the main release and reduce latency between upstream and the conda package.

Hi from oak park 👋

beckermr:
Hi from oak park too!

Yeah I am ok building them in parallel. 

You can also use a branch on the feedstock and only rebuild rcs periodically.

danpetry:
Just to clarify, I don't need rc builds on conda-forge's channel, personally. What I'm trying to do here is to align [Anaconda's recipe](https://github.com/AnacondaRecipes/pytorch-feedstock) with your recipe to foster closer collaboration and pool resources.  I spotted the point about rc builds in [this issue](https://github.com/conda-forge/pytorch-cpu-feedstock/issues/273) and thought this feature would be a good place to start.

With that said, happy to contribute the previous code suggestion if it fits your needs.

danpetry:
Are you thinking to make the rcs public? I think by the time we get an rc for a certain release building, the proper release will be pretty close if not already happened?

hmaarrfk:
> You can also use a branch on the feedstock and only rebuild rcs periodically.

Generally speaking this was my intention, somebody could keep a "1 commit PR" alive for referring to periodically.

trimmed down build matrix for reduced resource use and whatnot.

jakirkham:
Matt is referring to [CFEP 5]( https://github.com/conda-forge/cfep/blob/main/cfep-05.md ), which spells out how we handle RCs

It could be useful to have RCs to test that upstream changes build ok here or perhaps test combinations of packages locally

Typically we do a branch for RCs, but that sounds like a lot of effort to maintain here

Perhaps we could do the matrix approach and just add a `skip` that ignores the stable or RC build depending on the `version` specified. That way there should be no additional builds and it would be pretty easy to flip on/off

danpetry:
Something like this?

jakirkham:
Seems like a good starting point. Let's try [re-rendering]( https://conda-forge.org/docs/maintainer/updating_pkgs/#rerendering-feedstocks )

@conda-forge-admin , please re-render

danpetry:
This seems to work correctly for non-rc:
```
channel_targets:
- conda-forge main
...
rc:
- None
```
osx_64 is building.

Will switch to an rc build and see what the config and git fetching looks like.

jakirkham:
Also will want to re-render for the last change to take effect

danpetry:
@conda-forge-admin please re-render

danpetry:
thanks

danpetry:
I probably want to add something to the build string too

danpetry:
yep, that would be better

recipe/meta.yaml
```diff
@@ -1,4 +1,6 @@
 {% set version = "2.5.1" %}
+# Set to empty if not required, or rc<x> for rc version x
+{% set version_rc_suffix = "rc1" %}
```
jakirkham, line 3:
```suggestion
{% set version = "2.5.1rc1" %}
```

recipe/meta.yaml
```diff
@@ -17,11 +19,17 @@
 
 package:
   name: libtorch
-  version: {{ version }}
+  version: {{ version }}{{ version_rc_suffix }
```
jakirkham, line 22:
```suggestion
  version: {{ version }}
```

recipe/meta.yaml
```diff
@@ -17,11 +19,17 @@
...
 
 source:
+{% if rc != None %}
+  git_url: https://github.com/pytorch/pytorch.git
+  git_rev: v{{ version }}-{{ version_rc_suffix }}
```
jakirkham, line 27:
```suggestion
  git_rev: v{{ version.replace("rc", "-rc") }}
```

recipe/meta.yaml
```diff
@@ -56,6 +64,11 @@ build:
   # 2024/04 This build seems to fail due to disk space on the CIs
   # We can potentially re-enable it at a later time
   skip: true  # [cuda_compiler_version == "11.8" and aarch64]
+{% if version_rc_suffix %}
```
jakirkham, line 67:
```suggestion
{% if "rc" in version %}
```

jakirkham (commented):
Had some ideas on simplifications here

What do you think of this approach?

jakirkham:
This seems pretty good. Thanks for working on this Dan! 🙏

Would be interested to hear from others on this change 🙂

recipe/meta.yaml
```diff
@@ -20,8 +20,14 @@ package:
   version: {{ version }}
 
 source:
+{% if rc != None %}
+  git_url: https://github.com/pytorch/pytorch.git
+  git_rev: v{{ version.replace("rc", "-rc") }}
+{% else %}
```
h-vetinari, line 26: I'd prefer building from tags, not just https://github.com/pytorch/pytorch.git, e.g.

```suggestion
{% if rc != None %}
  url: https://github.com//pytorch/pytorch/archive/refs/tags/v{{ version.replace("rc", "-rc") }}.tar.gz
  sha256: <hash>
{% else %}
```
the downside however is that this wouldn't include submodules (due to git limitations) and pytorch uses [them](https://github.com/pytorch/pytorch/blob/v2.5.1-rc1/.gitmodules) _a lot_. So this is probably the better trade-off after all.

recipe/meta.yaml
```diff
@@ -1,4 +1,4 @@
-{% set version = "2.5.1" %}
+{% set version = "2.5.1rc1" %}
```
h-vetinari, line 1: I have a preference for making the `rc` bits a separate version component (but it works either way).
```suggestion
{% set version = "2.5.1.rc1" %}
```

recipe/meta.yaml
```diff
@@ -56,6 +62,11 @@ build:
   # 2024/04 This build seems to fail due to disk space on the CIs
   # We can potentially re-enable it at a later time
   skip: true  # [cuda_compiler_version == "11.8" and aarch64]
+{% if "rc" in version %}
+  skip: true  # [rc == "None"]
+{% else %}
+  skip: true  # [rc == "rc"]
+{% endif %}
```
h-vetinari, line 69: What's the point of this? We're only building one or the other anyway (depending on what's in `version`), so I'd just delete this (and the `rc:` variable in CBC) completely

recipe/conda_build_config.yaml
```diff
@@ -92,6 +92,17 @@ cuda_compiler_version:                              # [linux64]
...
+channel_targets:
+  - conda-forge main
+  - conda-forge pytorch_rc
+rc:
+  - None
+  - rc
+zip_keys:
+  -
+    - channel_targets
+    - rc
```
jakirkham, line 104: The `rc` bit is needed to select whether an RC label is used

recipe/meta.yaml
jakirkham: It is used to select whether and RC label is used for channel upload. Please see below

jakirkham (commented):
The channel upload is configured with RC like so

* `channel_targets` contains `main` and RC label
* `rc` is either `None` or `"rc"`
* `channel_targets` and `rc` are zipped together

When the recipe `version` contains `rc`...

* The case when `rc` is `None` is skipped
* The development source is selected
* Thus only the RC label for upload is kept

When the recipe `version` is stable

* The case when `rc` is `rc` is skipped
* The stable source is selected
* The `main` channel label is kept for upload

IOW this teases out the upload label

If there is a better way to do this, would be interested to hear 🙂

recipe/meta.yaml
danpetry: yes indeed. Are you ok with it as it is?

recipe/meta.yaml
danpetry: yes. Are you ok with this schema?

recipe/meta.yaml
h-vetinari: Yes, let's leave it as-is.

recipe/meta.yaml
danpetry: There's something I'm missing (after reading docs and related issues).

AFAIU:
- the `_numpy_rc` package is given the label `numpy_rc` in the conda-forge channel
- the pattern above means that if an rc build of numpy is desired, the `_numpy_rc` package, labelled `numpy_rc` is used
- due to strict channel priority, if the `numpy_rc` label is included as a higher-priority channel_source, any packages labelled `numpy_rc` are considered, and for those packages, versions/variants that aren't labelled `numpy_rc` are not considered.

I don't understand how this would affect channel_targets of numpy's rc builds though. Does channel_targets automatically change to `conda-forge numpy_rc` if `_numpy_rc` is pulled, somehow? Also, I can't see any numpy packages labelled `numpy_rc` on the [conda-forge channel](https://anaconda.org/conda-forge/numpy/labels), and I can see rc versions being uploaded [without a `numpy_rc` label](https://anaconda.org/conda-forge/numpy/files).

danpetry: I also don't see how deleting the `rc` variable in cbc.yaml would work, if we want to have two channel_targets (which would need to be zipped with something, I guess, otherwise everything would be built twice and uploaded to both labels)

danpetry: I'm still getting familiar with conda-forge's CI/smithy/rendering etc, so thanks for any info!

danpetry:
@conda-forge-admin please re-render

recipe/meta.yaml
h-vetinari: > I don't understand how this would affect channel_targets of numpy's rc builds though. 

That's the neat trick about this. They can be published to `main`, and only the "magic" package `_numpy_rc` needs to be under a different channel  (under the opt-in `conda-forge/label/numpy_rc`)
 
> Does channel_targets automatically change to `conda-forge numpy_rc` if `_numpy_rc` is pulled, somehow? Also, I can't see any numpy packages labelled `numpy_rc` on the [conda-forge channel](https://anaconda.org/conda-forge/numpy/labels), and I can see rc versions being uploaded [without a `numpy_rc` label](https://anaconda.org/conda-forge/numpy/files).

How this works is that:
```
conda install numpy=2.2.0rc1  # fails due to missing dependency _numpy_rc
conda install -c conda-forge/label/numpy_rc numpy=2.2.0rc1  # succeeds
```

This isn't necessary per se for pytorch, but it's a less intrusive approach to handle rc-builds IMO (at the cost of having an extra magic package).

isuruf: > This isn't necessary per se for pytorch, but it's a less intrusive approach to handle rc-builds IMO (at the cost of having an extra magic package).

Let's not do this for all packages. The accepted behaviour is to put the RC package into `conda-forge/label/pytorch_rc` as specified in the CFEP. The magic package behaviour is to work-around some limitations with conda-forge CI infrastrcture when building new python/numpy packages based on the RCs. Since we are not going to build downstreams with one released version and one rc, this is not needed and adds extra complexity.

h-vetinari: Fine by me!

danpetry: ok, thanks for the explanation, I understand that now. It stops RCs from being installed, unless you specify that label with `-c`.
@h-vetinari @jakirkham are you good with this as-is? Any further comments?

danpetry: Apart from removing the `rc1` in the version string

h-vetinari: > are you good with this as-is? Any further comments?

The main remaining question to me is whether we're going to be building rc's so often that it's justified to add this infrastructure to `main`. If the answer is yes: fine by me. Historically, this feedstock has moved slowly enough that rc's were built (if at all) in PRs, and could then turn into a regular version bump as soon as the version went GA.

danpetry: I mentioned this at the start of the PR:

> What are your requirements around this in general? We use it to prepare for the main release and reduce latency between upstream and the conda package. We don't upload the rc packages.

Since I'm pretty new to understanding conda-forge culture/requirements, I went with the guidance from @hmaarrfk and @jakirkham . Assuming that what they were asking reflected the intention behind the first bullet point [here](https://github.com/conda-forge/pytorch-cpu-feedstock/issues/273).

jakirkham: Personally am good with what we have here

Agree with Isuru that we likely don't need the `pytorch_rc`  package approach here. That would only be useful if we want to migrate to a PyTorch RC in conda-forge and we think there will be enough time between the RC and the final release to merit that effort. Personally am not convinced we need this additional complexity. Could be convinced if we start packaging RCs more regularly and quickly

My understanding is this addition of RC builds was helpful also for alignment with Anaconda's package. So tried to recommend a minimal solution to accomplish that goal

Think it would be good to hear from Mark on these changes

jakirkham:
To fix conflicts, would recommend...

1. Check this out locally
2. Attempt merge
3. Resolve conflicts on `recipe/` only
4. `git rm -f '.ci_support/linux_*' '.ci_support/osx_*' '.ci_support/win_*'`
5. [Re-render]( https://conda-forge.org/docs/maintainer/updating_pkgs/#rerendering-feedstocks )
6. Finish merge and push changes

recipe/meta.yaml
danpetry: I agree. Additional to what we will pull across, this includes a minimal mechanism to stop rc builds being used by normal users too. So seems minimal as-is.
@hmaarrfk any comments?

danpetry:
Thanks. Any particular reason you recommended to git rm the CI files (which would be re-generated upon render anyway), and also to render before finishing merge? I'm guessing, just because it's less likely to run into messy git errors, but I wanted to make sure that's all

recipe/meta.yaml
hmaarrfk: I mostly don't want to give "an other" opinion.

I think that `rc` builds are generally helpful to anticipate build problems.


The additions here seem "minimal" and well thought out.

it also gives us the ability to give feedback to upstream 

But with 2Month+ merge times on 1 line fixes:
* https://github.com/pytorch/pytorch/pull/137331
* https://github.com/pytorch/pytorch/pull/137084

I'm personally not so enthusiastic about finding "suggestions" during the `rc` and having them merged in before the full release.

Ultimately, I'm very glad to see some new contributors to this feedstock. If you think the `rc` builds would empower you to continue to contribute (perhaps by using this feedstock and uploading to your own channel) that is enough for me to accept this in.

We can always "remove" the `rc` stuff later.

jakirkham:
Yeah that's basically it. Also skips spending time on conflicts that will go away with a re-render

recipe/meta.yaml
danpetry: It's less about providing upstream fixes, and more about getting the feedstock to build ahead of the release date, so that it's available to conda users with minimal latency compared to the pypi package.
Which is especially useful now that pytorch have dropped conda packages.

danpetry: I basically want to merge anaconda's recipe with this one a bit more. And thought this would be a relatively trivial, easy-first-issue sort of change 😆

danpetry: I'm learning that conda-forge's recipes, being public, need to be gone over more than Anaconda's ones, because they'll be used by third parties and need to be more foolproof

hmaarrfk: basically for me you've made it pretty clear:

> danpetry is willing to help. he has taken the initiative to work through concerns of adding an `rc` and shown to be receptive to feedback. I think this is worth a try to help unify the recipe and get him invested in the maintenance.


Regarding:

> I'm learning that conda-forge's recipes, being public, need to be gone over more than Anaconda's ones

Its mostly that this recipe is very popular by end users. and it is performance critical. Many core members have used it in their work (myself included) so its just a little more touchy.

We also require many special resources, CFEP03 or self-hosted runners, which start to necessitate coordination.

danpetry: My intention is that in the future we can pool resources rather than develop this recipe separately

jakirkham:
Do we need to start a new cirun build here?

h-vetinari:
> Do we need to start a new cirun build here?

#298  is a big PR that's almost complete; I'm going to ask that we prioritize merging that one first.

h-vetinari:
It doesn't make it easier, because that PR is a major reshuffle and we shouldn't also layer conflicts with unrelated changes on top (aside from the fact that the CI here is deep red). This PR is more focussed, and will be easier to adapt than vice versa. I can offer to do it even.

danpetry:
I am fine to wait for that other PR. I'm also more or less on Christmas break now until January, so I'm fine to wait.

jakirkham:
Given #298 is now in, can we wrap this one up and merge?

Note CI would pass if someone approved cirun on this PR. Idk that we need it though given the changes already passed on CPU only jobs and nothing here is GPU specific

danpetry:
Has been re-rendered

recipe/meta.yaml
```diff
@@ -56,6 +63,11 @@ build:
   skip: true  # [win]
   # cuda 11.8 was dropped due to maintenance effort, see discussion in #177
   skip: true  # [cuda_compiler_version == "11.8"]
+{% if "rc" in version %}
+  skip: true  # [rc == "None"]
+{% else %}
+  skip: true  # [rc == "rc"]
+{% endif %}
```
h-vetinari, line 70: Correct me if I'm wrong, but I don't see how anything would be built for `2.6.0.rc1`) unless we fully duplicate our variant configs (to have variants for both values of `rc`). Currently, this PR adds
```yaml
rc:
- None
```
to all of them, but that means we'd just skip all builds if someone adds `.rcX` to the version.

h-vetinari (changes requested):
I don't think this is functional yet (see below).

As an aside, I think the variable/values are poorly named, if it causes us to write `[rc == "rc"]`. AFAICT we could just do
```yaml
rc:
  - True
  - False
```
and then use `# [not rc]` and `# [rc]`.

I'm not excited about the duplication of the variant configs, but if people thing this is a net benefit, then so be it.

.ci_support/osx_64_blas_implgenericchannel_targetsconda-forge_mainnumpy2.0python3.12.____cpythonrcNone.yaml
```diff
@@ -52,10 +52,14 @@ python:
 - 3.12.* *_cpython
 pytorch:
 - '2.4'
+rc:
+- None
 target_platform:
 - osx-64
```
jakirkham, line 58: Here is the macOS `x86_64` build for Python 3.12 with `rc` set to `None`

The corresponding [Azure CI build]( https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=1137925&view=logs&jobId=c8b8640e-759f-5af3-65b5-809c078c71b1&j=899534ae-d486-52b7-7b1c-5c9a77433eb6&t=c641a15e-a70f-5a9e-1db4-6b8087a1df4f ) ran for ~5hrs

So AFAICT this is working as expected

recipe/meta.yaml
jakirkham: When one adds an `rc` version, they will need to re-render. Doing so updates the value like so

```yaml
rc:
- rc
```

.ci_support/osx_64_blas_implgenericchannel_targetsconda-forge_mainnumpy2.0python3.12.____cpythonrcNone.yaml
h-vetinari: That's because the version here currently does not contain `"rc"`. Switch to the 2.6.0.rc3 builds, and you'll see (if I'm right) that everything will get skipped.

jakirkham: Please see my comment above: https://github.com/conda-forge/pytorch-cpu-feedstock/pull/280#discussion_r1906676886

recipe/meta.yaml
h-vetinari: OK, now I finally understand the intended mechanics, thanks.

recipe/conda_build_config.yaml
```diff
@@ -4,6 +4,17 @@ c_stdlib_version:  # [unix]
 MACOSX_SDK_VERSION:         # [osx]
   - 13.3                    # [osx]
 
+channel_targets:
+  - conda-forge main
+  - conda-forge pytorch_rc
+rc:
+  - None
+  - rc
```
h-vetinari, line 12: Please change this to
```suggestion
is_rc:
  - False
  - True
```

h-vetinari (changes requested):
OK, I understood the mechanics now, pretty clever. :)

I've played around with this on a [branch](https://github.com/h-vetinari/pytorch-cpu-feedstock/tree/rc-builds) and it works as expected. However, please follow the rename below (or simply pick https://github.com/h-vetinari/pytorch-cpu-feedstock/commit/8f6694a2a02fce041537f2448ebb0a6646a48e32)

jakirkham:
FWIW commit ( https://github.com/h-vetinari/pytorch-cpu-feedstock/commit/8f6694a2a02fce041537f2448ebb0a6646a48e32 ) looks reasonable to me

danpetry:
that's better, thanks

danpetry:
I've rendered locally but want to see how it goes in the CI, for rc and non-rc

recipe/meta.yaml
```diff
@@ -1,4 +1,5 @@
-{% set version = "2.5.1" %}
+# if you wish to build release candidate number X, append the version string with ".rcX"
+{% set version = "2.5.1.rc1" %}
```
jakirkham, line 2:
```suggestion
{% set version = "2.5.1" %}
```

jakirkham (changes requested):
Thanks Daniel! 🙏

The changes look good

Though let's make sure to go back to a stable version (and re-render) before merging

h-vetinari:
I've picked the changes from this PR into https://github.com/conda-forge/pytorch-cpu-feedstock/pull/305/commits/63779a3ca38c50fe351da9ecd05a952e9360a4a7, given that 30 commits are a bit much for the core change of
```
 2 files changed, 26 insertions(+)
```
😉 

Planning to merge this with #305 as soon as CI passes there.

danpetry:
Cool sounds good. Closing this one, then.

recipe/meta.yaml
danpetry: yup

jakirkham:
Thanks you both and everyone who helped review! 🙏

danpetry:
indeed thanks 🎉

PR #289, opened 2024-11-08
Use system mkldnn/onednn

mgorny:
Checklist

-------
Patch pytorch to use the sysem mkldnn library from the onednn package rather than building one locally from within ideep submodules.  Given that ideep itself is a header-only library, I presume this is what was meant in https://github.com/conda-forge/pytorch-cpu-feedstock/issues/108#issuecomment-2398442313, and indeed unvendoring onednn seems to improve build time significantly.

That said, our onednn package does not support GPU runtime (https://github.com/conda-forge/onednn-feedstock/pull/44) but at least according to my testing, that part of the library was not enabled by our PyTorch builds before (due to missing SYCL).

The patch is a bit hacky, and probably needs some polishing before being submitted upstream (and testing on other platforms).

Part of issue #108


beckermr (dismissed):
I am confused by this PR. We need any libraries linked by pytorch to be provided by conda-forge.

mgorny:
> I am confused by this PR. We need any libraries linked by pytorch to be provided by conda-forge.

`onednn` is provided by conda-forge. For some (historical?) reason it's named `mkldnn` in PyTorch CMake files.

beckermr:
Ahh I see. LGTM!

mgorny:
Hmm, I also should make the `onednn` dependency conditional. Need to test it a little more too.

mgorny:
This will also require a patch to onednn-feedstock, to enable experimental ukernel. I will submit it once I finish testing (probably next week).

hmaarrfk:
You should keep the feature set consistent between windows/unix if you can. 

* [ ] Modify the .bat script as well

You might need to disable things for non-x86_64 arch. You can test `target_platform` in the sh file.

hmaarrfk:
Lets see how the improvement translates to azure.

recipe/patches/0009-Use-system-mkldnn.patch
```diff
@@ -0,0 +1,51 @@
+diff --git a/CMakeLists.txt b/CMakeLists.txt
+index 98593c2d..94a9d63d 100644
+--- a/CMakeLists.txt
++++ b/CMakeLists.txt
```
h-vetinari, line 4: For quickly applying the patches as they exist on the feedstock (i.e. if it's not you who did the last change to them), it's nicer to generate them with `git format-patch <tag> --no-signature`.

That has several advantages:
* commit message can be used to provide context (this patch is pretty self-explanatory, but in general...)
* authorship is easily visible (who to ask in case there are questions about intent, e.g. when rebasing to new versions)
* easily apply them to a source-checkout in one go using e.g.
  `find ../path/to/feedstock/recipe/patches/ | xargs git am`
* having a local branch of patches makes it much easier to rebase the changes to a new version
* etc.

So while raw diff's are technically possible, the output of `git format-patch` is preferred.

h-vetinari (commented):
The direction here looks very good (custom channel aside of course). 👍

Issue #233, opened 2024-04-09
Build with mkldnn support alongside cuda


shunr:
### Comment:

Hi,

I'm currently working on torch code that should run on both cpu and cuda backends. Specifically, I'm using `bfloat16` which requires the mkldnn backend to be performant on cpu. Is it possible to ship the cuda build with mkldnn turned on?

https://github.com/conda-forge/pytorch-cpu-feedstock/blob/9e99e0322b67ac080cb315cd1bb2b9b9e6bb9f15/recipe/build.sh#L121-L123
The above comment seems to suggest that USE_MKLDNN=1 should in theory be usable with cuda, but some compilation error led to a temporary workaround.

Thanks.

mgorny:
FYI, I've been able to enable it successfully *after* applying the fixes from #291.

PR #291, opened 2024-11-18
Stop passing CMAKE_TOOLCHAIN_FILE, append to CMAKE_FIND_ROOT_PATH

mgorny:
Checklist

-----------------

    Stop passing obsolete CMAKE_TOOLCHAIN_FILE for CPU builds.  Instead,
    append the source directory to CMAKE_FIND_ROOT_PATH directly.  Do that
    unconditionally, to improve consistency between CPU and GPU builds,
    and hopefully make it possible to build MKLDNN with CUDA enabled.

---------------

I haven't bumped the build number since this shouldn't be changing anything (and at least at a first glance, I didn't notice any differences). It's getting late today, so I'm going to look at whether this actually fixes building against vendored mkldnn tomorrow (I've originally tested only against my system-mkldnn branch).


mgorny:
And thanks to @isuruf for explaining the toolchain thing to me.

hmaarrfk:
> I haven't bumped the build number

If this is the case, I would rather you set the time limit to 1 min to effectively force the CI to not publish at all. I am a little concerned about the GPU build since it doesn't show up on the CIs.

mgorny:
> > I haven't bumped the build number
> 
> If this is the case, I would rather you set the time limit to 1 min to effectively force the CI to not publish at all. I am a little concerned about the GPU build since it doesn't show up on the CIs.

I'm sorry, I didn't know I can do that. In fact, I haven't found any information on that yet. Perhaps I could use `[ci skip] ***NO_CI***` in the future instead?

Also, I'm currently confirming that this fixes `USE_MKLDNN=1` with CUDA, per #233. Would you like me to add enabling it to this pull request, or wait for it till we unvendor it (per #289) instead?

Issue #177, opened 2023-07-31
Enable other BLAS implementations with CUDA


jakirkham:
Recently there was some work in PR ( https://github.com/conda-forge/pytorch-cpu-feedstock/pull/175 ) to enable different BLAS implementations to be used with PyTorch in the CPU-only case. Am curious if this can be extended to the CUDA case

Has this been explored before? If so, would be interested to learn what those past attempts learned. If not, would be curious to know if there are issues that would need to be addressed first

mgorny:
Unless I'm mistaken, it's technically possible to use CUDA + generic BLAS. I've just tried and it builds just fine (just had to remove the respective skips).

That said, I think the main problem is that allowing a second BLAS implementation would double the number of variants. And the CUDA variants are both very slow to build and are quite big. I think that the first problem can be somewhat circumvented using ccache (at least during my testing I didn't notice the second build rebuilding CUDA objects again).

If @hmaarrfk is okay with these, I can make a PR to stop skipping them; either for all CUDA variants or a subset of them.

hmaarrfk:
I would likely just drop 11.8 if it were up to me. 

Ubuntu all the way back to 20.04 enables installation of cuda 12. 

If you want to try to add caching that is fine too, but historically it has been more pain than not.

The second blas variant has been a long requested feature so I'm all in!

mgorny:
> I would likely just drop 11.8 if it were up to me.

Ok, I'll put up a PR with that + enabling generic BLAS (tomorrow, I think), and let's see what others think.

> If you want to try to add caching that is fine too, but historically it has been more pain than not.

Well, I've only succeeded while using it locally while calling `conda build` directly. I honestly have no clue how to do it with Docker — probably will need wider ecosystem changes.

jakirkham:
What are the MKL bits used here for?

If it is just MKL-DNN, perhaps that makes less sense in the CUDA builds, in which case we could simply switch to generic BLAS in that case. That would avoid an additional build job.

Alternatively if it is straightforward to identify and pull in the MKL bits that are needed outside of BLAS and LAPACK, it may be possible to always use a generic BLAS and pull in a more focused portion of MKL

hmaarrfk:
There were apparently some uses of FFT routines.

mgorny:
> What are the MKL bits used here for?

[...]

PR #292, opened 2024-11-22
Allow "generic" BLAS builds with CUDA enabled

mgorny:
Checklist

----
Reenable generic BLAS when CUDA is used (#177), and disable CUDA 11.8 while at it, as discussed in that bug.


recipe/meta.yaml
```diff
@@ -52,7 +52,6 @@ build:
     - python *                               # [megabuild]
     - numpy *                                # [megabuild]
   skip: true  # [win]
-  skip: true  # [cuda_compiler_version != "None" and linux64 and blas_impl != "mkl"]
```
h-vetinari, line 55: since that commit came after the rerender, you didn't actually create any new jobs?

Also, for the time we're debugging this, you might want to inverse this condition (only build CUDA+non-MKL) to save CI resources.

mgorny: This doesn't affect available jobs — they were always there, they just terminate without building or producing any artifacts. And yes, that confused me too, so I've checked.

h-vetinari: Yeah, what happens is that some more complicated skip conditions don't get detected correctly by smithy (or are hard to specify so that this happens), which I guess is what happened here.

hmaarrfk:
have you confirmed that it works locally on a larger machine with docker? Using the python build locally py script?

If so, Can you move to cirun so that we can run the Linux builds on the CI?

mgorny:
> have you confirmed that it works locally on a larger machine with docker? Using the python build locally py script?

No, I've built it using `conda build` directly, since I had a warm ccache for it. Building from scratch via Docker would take a few hours for me.

Issue #283, opened 2024-11-04
Investigate being unable to find cublas_v2.h


jeongseok-meta:
https://github.com/pytorch/pytorch/blob/v2.5.1/cmake/Modules/FindCUDAToolkit.cmake#L809

```
  -- Unable to find cublas_v2.h in either "/home/conda/feedstock_root/build_artifacts/libtorch_1730670943607/_build_env/targets/sbsa-linux/include" or "/home/conda/feedstock_root/build_artifacts/libtorch_1730670943607/_build_env/math_libs/include"
```

Affected builds:
- linux64 + mkl + cuda 12.0/12.6
- aarch64 + cuda 12.0/12.6

Related:
- https://github.com/pytorch/pytorch/issues/132599

_Originally posted by @hmaarrfk in https://github.com/conda-forge/pytorch-cpu-feedstock/issues/271#issuecomment-2453640343_

hmaarrfk:
this was in the aarch64 + cuda 12.6 build log.

jeongseok-meta:
FWIW, this issue is not new, as it also occurred in a build log from two weeks ago (randomly picked): https://github.com/conda-forge/pytorch-cpu-feedstock/actions/runs/11317968091/job/31472007572

hmaarrfk:
It seems like it is available in the `libcublas-dev` package since at least 12.0
https://conda-metadata-app.streamlit.app/?q=conda-forge%2Flinux-64%2Flibcublas-dev-12.0.1.189-hd3aeb46_3.conda

jakirkham:
It is worth noting the paths in the error are from the `build` environment and not the `host` environment. Do we know why it isn't using the latter?

mgorny:
Indeed looks like CMake determines CUDA path based on `nvcc` location, and our overrides don't seem to work. I'm going to try figuring out if we can override this properly without patching, or prepare a patch to fix it.

mgorny:
Good news, everyone! It seems that I have a fix, it wasn't as horrible as I've anticipated, and additionally it seems to remove the need for build dependencies on the additional CUDA libraries. Not creating a PR yet, as I don't want to torture the CI: https://github.com/mgorny/pytorch-cpu-feedstock/compare/cufile...cuda-prefix

PR #296, opened 2024-11-29
Next batch of changes (notably, CUDA prefix fixes)

mgorny:
Checklist

---------------

1. Removed code for building against old CUDA versions, and synced the Windows `TORCH_CUDA_ARCH_LIST` for 12.6.
2. Patch and fix the PyTorch build system to use CUDA libraries from PREFIX, fixing finding cublas_v2 #283.
3. Remove the hack adding CUDA deps to build — I think it was added because of the wrong prefix above, and I was able to build it successfully after removing this.
4. Unvendor six — it was implicitly downloaded by NNPACK build system.
5. Remove unused dependencies on future and typing.
6. Added a rough list of dependencies used upstream (vendored / used in wheels) vs. ours, as well as info where to find them.


recipe/README.md
```diff
@@ -46,3 +48,54 @@ done
...
+...
+```
+
+Whenever possible, it will describe the used commit using a version tag.
+
+For Python dependencies, the specific pins can be found
+in `*.dist-info/METADATA` inside the wheel. For non-Python dependencies,
+CI scripts can be inspected.
+
+The following table summarizes versions used in 2.5.1 (`+` indicates
```
h-vetinari, line 76: This is nice. It'll get out of date pretty quickly, but good to see. :)

h-vetinari (commented):
This looks very nice already 👌

I'm going to push a small fix to trigger a build.

h-vetinari:
> I'm going to push a small fix to trigger a build.

Ah sorry I didn't see your push @hmaarrfk, it came while I was reviewing!

h-vetinari (approved):
👍

recipe/patches/0010-Allow-overriding-CUDA-related-paths.patch
```diff
@@ -0,0 +1,39 @@
...
+index ec9ae530..76d1fb9a 100644
+--- a/cmake/Modules/FindCUDAToolkit.cmake
++++ b/cmake/Modules/FindCUDAToolkit.cmake
+@@ -497,7 +497,7 @@ Result variables
+ # CMAKE_CUDA_COMPILER_TOOLKIT_ROOT and CMAKE_CUDA_COMPILER_LIBRARY_ROOT.
+ # We compute the rest based on those here to avoid re-searching and to avoid finding a possibly
+ # different installation.
+-if(CMAKE_CUDA_COMPILER_TOOLKIT_ROOT)
++if(FALSE)
+   set(CUDAToolkit_ROOT_DIR "${CMAKE_CUDA_COMPILER_TOOLKIT_ROOT}")
```
hmaarrfk, line 21: is there a way to make this patch "upstreamable"? While they have been slow to merge patches they do work to merge them eventually....

hmaarrfk (approved):

recipe/patches/0010-Allow-overriding-CUDA-related-paths.patch
mgorny: That's a very good question. I suppose the cleanest approach would be to add a new variable to control/override this behavior — though I would find it hard to justify it beyond "this is the simplest way I managed to make the current deeply complex logic let me use a separate 'host' tree for conda".

hmaarrfk: It’s funny because I literally wrote the same message elsewhere. 

Maybe we can open an internal tracking issue to discuss longer term solutions. Upstream has historically periodically compared their build systems to ours. But it is unclear how this is evolving as they dropped conda support.

mgorny: I think the biggest problem here is that we're seeing an intersection of three different sets of logic:

1. Built-in CMake support for using the CUDA compiler.
2. The `FindCUDA` module with its patched version inside PyTorch.
3. The `FindCudaToolkit` module within PyTorch.

I think what upstream is really trying to do here is gluing they all together, so they don't end up using different CUDA installations, but I don't know how many different use cases they are trying to cover here. On top of that, I think Conda is a bit of an "odd one" here, using separate build and host trees even when not actually cross-compiling.

Issue #180, opened 2023-08-04
`torch.dot()` produce wrong results


husisy:
### Solution to issue cannot be found in the documentation.

- [X] I checked the documentation.

### Issue

claim: `pip install torch` gives the correct results, so I think it should be an issue to be reported here.

claim: right now (2023-08-04) i can only reproduce this issue on macOS (I test both M1 and M2 silicon), cannot reproduce on ubuntu-22.04 with cuda.

minimum working example (MWE) to reproduce wrong results

```bash
## use conda/micromamba either ONE

conda create -y -n test00
conda install -y -n test00 -c conda-forge python pytorch
conda activate test00

## conda-forge is the default one (usually)
micromamba create -y -n test00
micromamba install -y -n test00 python pytorch
micromamba activate test00
```

```python
# draft00.py
import torch
x = torch.tensor([2], dtype=torch.complex128)
print(torch.dot(x, x))
```

```bash
python draft00.py
# tensor(0.+0.j, dtype=torch.complex128) ##should be 4
```

MWE to reproduce correct results

```bash
# similar for micromamba
conda create -y -n test01
conda install -y -n test01 -c conda-forge python numpy
# if no numpy, UserWarning is raised
conda activate test01
pip install torch
```

```python
# draft00.py
import torch
x = torch.tensor([2], dtype=torch.complex128)
print(torch.dot(x, x))
```

```bash
python draft00.py
# tensor(4.+0.j, dtype=torch.complex128)
```

### Installed packages

```shell
# packages in environment at /Users/zhangc/miniconda3/envs/test00:
#
# Name                    Version                   Build  Channel
bzip2                     1.0.8                h3422bc3_4    conda-forge
ca-certificates           2023.7.22            hf0a4a13_0    conda-forge
filelock                  3.12.2             pyhd8ed1ab_0    conda-forge
gmp                       6.2.1                h9f76cd9_0    conda-forge
gmpy2                     2.1.2           py311h2ba9262_1    conda-forge
jinja2                    3.1.2              pyhd8ed1ab_1    conda-forge
libblas                   3.9.0           17_osxarm64_openblas    conda-forge
libcblas                  3.9.0           17_osxarm64_openblas    conda-forge
libcxx                    16.0.6               h4653b0c_0    conda-forge
libexpat                  2.5.0                hb7217d7_1    conda-forge
libffi                    3.4.2                h3422bc3_5    conda-forge
[...]

PR #298, opened 2024-12-04
Recipe overhaul: more tests & documentation, various clean-ups

### LINKED ISSUE #275: Make Magma optional for cuda builds?
### Comment:

A conda-forge environment with nothing but pytorch for cuda in it currently ways in at 7.2GB, which is perceived to be rather on the heavy side of things.

Looking at potential for slimming things down, libmagma at ~2GB looks like a good candidate.

The pytorch docs [seem to suggest](https://pytorch.org/docs/stable/backends.html#torch.backends.cuda.preferred_linalg_library) that libmagma is used as an alternative for cusolver, which is included anyway, at a much more modest 150 MB.

In the past, magma was significantly faster than cusolver, as demonstrated by [1]. However, a recent 2024 paper by the magma authors [2] shows that cusolver has made progress and is now faster for some of the most important problems.
Magma still offers significant performance benefits for certain workloads, but given that pytorch has the ability to switch between the available libraries, we could make magma an optional dependency, i.e. merely include it in `run_constrained` and leave it up to the user to choose space or performance optimization based on their use-case.

Is this feasible or am I missing something about the use of magma in pytorch?

Do you think this is desirable?

[1] S. Abdelfattah, A. Haidar, S. Tomov and J. Dongarra, "Analysis and Design Techniques towards High-Performance and Energy-Efficient Dense Linear Solvers on GPUs," in IEEE Transactions on Parallel and Distributed Systems, vol. 29, no. 12, pp. 2700-2712, 1 Dec. 2018, doi: 10.1109/TPDS.2018.2842785. keywords: {Graphics processing units;Energy efficiency;Task analysis;Multicore processing;Dense linear solvers;GPU computing;energy efficiency},

[2] Abdelfattah A, Beams N, Carson R, et al. MAGMA: Enabling exascale performance with accelerated BLAS and LAPACK for diverse GPU architectures. The International Journal of High Performance Computing Applications. 2024;38(5):468-490. doi:10.1177/10943420241261960

mgorny:
Checklist

<details><summary>Old description until Dec 23, 2024</summary>

From the commit message:

    Upstream keeps all magma-related routines in a separate
    libtorch_cuda_linalg library that is loaded dynamically whenever linalg
    functions are used.  Given the library is relatively small, splitting it
    makes it possible to provide "magma" and "nomagma" variants that can
    be alternated between.

Also:

    Try to speed up magma/nomagma builds a bit.  Rather than rebuilding
    the package 3 times (possibly switching magma → nomagma → magma again),
    build it twice at the very beginning and store the built files for later
    reuse in subpackage builds.
    
    While at it, replace the `pip wheel` calls with `setup.py build` to
    avoid unnecessarily zipping up and then unpacking the whole thing.
    In the end, we are only grabbing a handful of files for `libtorch*`
    packages and they are in predictable location in the build directory.
    `pip install` remains being used in the final builds for `pytorch`.

In this PR's implementation, we've chosen to prioritize the magma build for those with GPUs. We have done so by using both `track_features` on nomagma, and an increased build number for the magma build with should help different solvers naturally find the magma build.

Fixes #275 
</details>

Due to the desire to decrease the maintenance burden, and the fact that we were able to reduce the compiled size of the libmagma package, we decided to abandon the original goal of making magma optional.

Instead this PR has changed focus to a refactor of the original recipe bringing tests and documentations into light.

@isuruf helped me a lot with this, particularly with refactoring the builds, so both variants are built in one run.


h-vetinari:
Since this is still WIP, I started only a single job on linux for now (x64+MKL+CUDA)

mgorny:
> Since this is still WIP, I started only a single job on linux for now (x64+MKL+CUDA)

Thanks. It seems to have failed only because it's adding new outputs. Do you want me to file the admin request for allowing `libtorch-cuda-linalg`, or do you prefer reviewing the changes first?

h-vetinari:
> Do you want me to file the admin request for allowing `libtorch-cuda-linalg`

That would be great!

mgorny:
Filed as https://github.com/conda-forge/admin-requests/pull/1209.

recipe/README.md
```diff
@@ -98,3 +98,86 @@ of 2024-11-28:
...
+
+Packages built by the recipe
+----------------------------
+The recipe currently builds four packages:
+
+1. `libtorch` that installs the common libraries, executables and data files
+   that are independent of selected Python version and are therefore shared
+   by all Python versions.
+
+2. `libtorch-cuda-linalg` that provides the shared `libtorch_cuda_linalg.so`
```
hmaarrfk, line 114: is this the preffered one?

as in, is it preffered that users make use of linalg?

I ask because we try very hard to ensure that

```
mamba install pytorch
```
installs the best hardware optimized one.

mgorny: Not sure I understand the question.

`libtorch_cuda_linalg.so` is required for some operations. It is always pulled in by `libtorch` itself, i.e. the end result is that some version of the library is always installed.

As for magma vs. nomagma, I think we ought to prefer magma. https://github.com/conda-forge/pytorch-cpu-feedstock/issues/275#issue-2588203757 suggests that cusolver will be faster for other workflows, but the magma build supports both and according to https://pytorch.org/docs/stable/backends.html#torch.backends.cuda.preferred_linalg_library, it has a heuristic to choose the faster backend for given operation.

hmaarrfk: So who benefits from making this optional? 

Sorry if this obvious.

mgorny: This is #298, i.e. people who want to avoid installing the large magma dependency (~250M in package). Given that `libtorch-cuda-linalg` is 200k (nomagma) / 245k (magma), I've figured out it's worth it.

mgorny:
I've added explanation in the README, as well as a fix to install `libtorch_python` symlink, and missing test dependencies. Tests don't work yet but I'm on a good way to run a subset of them (like upstream CI does).

recipe/README.md
rgommers: The no-magma variant reduces the on-disk install size of a `pytorch-gpu` install from 7 GB to 5 GB (as #275 says). So this is definitely worth doing. 

Given that `cusolver` is now on average fairly close in performance to `magma` _and_ that the PyTorch wheels don't include magma, I'd argue that the default should be `nomagma` (I'd certainly almost always want that for my use cases). However, what the default is is less important than both options being available.

rgommers: I'd also explain the size difference explicitly in the README, so it's clear what the tradeoff is.

hmaarrfk: ok this makes sense, thanks for the in depth discussion.

I am weary of "following upstream". So instead of "this is what upstream does on pypi", I want to make sure we are asking

1. Do we believe that magma usage is niche enough to omit by default and should be an opt in.

it seems that @mgorny thinks we should make it opt out, while @rgommers thinks we should make it opt in. I'll let you all decide, but hopefully you can document your intentions.

Over the years, we've just gone through many different ideas (track features, build numbers, etc....) to help users get a good default experience.

mgorny: I don't really have a strong opinion on it. My suggestion was basically "minimal change", but I'm fine with either default.

rgommers: > Pytorch statically links to magma. 

Argh, I missed that, hard to tell from an unpacked wheel. 

So is the 2 GB size difference explained by the Magma shared libraries being that large but PyTorch only needing a small fraction of the functionality in Magma, or is there more to it?

mgorny: Well, I don't know about the latter point, but yes, indeed, `libmagma` and `libmagma_sparse` take 2G:

```
$ du -hs $CONDA_PREFIX
7.4G	/home/mgorny/.conda/envs/pytorch
$ conda uninstall --force libmagma libmagma_sparse

## Package Plan ##

  environment location: /home/mgorny/.conda/envs/pytorch

  removed specs:
    - libmagma
    - libmagma_sparse


The following packages will be REMOVED:

  libmagma-2.8.0-h0af6554_0
  libmagma_sparse-2.8.0-h0af6554_0


Proceed ([y]/n)? y


Downloading and Extracting Packages:

Preparing transaction: done
Verifying transaction: done
Executing transaction: done
$ du -hs $CONDA_PREFIX
5.4G	/home/mgorny/.conda/envs/pytorch
```

h-vetinari: `libmagma` itself is [around](https://anaconda.org/conda-forge/libmagma/files) 250MB, and `libmagma_sparse` [<10MB](https://anaconda.org/conda-forge/libmagma_sparse/files). So the vast majority of that footprint comes through transitive dependencies (which might also be depended on by something else in the environment, erasing the gains).

The `libmagma` [build](https://conda-metadata-app.streamlit.app/?q=conda-forge%2Flinux-64%2Flibmagma-2.8.0-hfdb99dd_0.conda) for CUDA 11.8 depends on `cudatoolkit`, which is pretty hefty, but then, the CUDA 11.8 builds are fast becoming obsolete. The CUDA 12.0+ [build](https://conda-metadata-app.streamlit.app/?q=conda-forge%2Flinux-64%2Flibmagma-2.8.0-h0af6554_0.conda) mainly depends on `libcublas` (~260MB) and `libcusparse` (~120MB).

The rest of the dependencies are either tiny or ubiquitous (libblas, libgcc, libstdcxx), so I don't think the 2GB footprint is realistic. It sounds more like ~650MB to me. Which is still big (so no strong opinion whether we want it by default), but two thirds less than 2GB.

mgorny: My numbers were actually for installed artifacts, i.e. literally:

```
-rwxrwxr-x 2 mgorny mgorny 2.0G Jul 22 22:55 /home/mgorny/.conda/envs/pytorch/lib/libmagma.so
```

h-vetinari: > My numbers were actually for installed artifacts, i.e. literally:

Ok, thanks. Better compression ratio than I expected! Perhaps as an additional data point, could we compare the total unpacked footprint for a pytorch environment with/without magma?

2GB is huge in absolute terms, but perhaps less so if we're talking relatively (say if it's ~10-15%).

I'd do it myself but currently travelling and on a metered connection, so too big to try right now

h-vetinari: Thanks. That certainly supports the argument that it should be opt-in (I guess as another alternative we could consider packaging static libmagma and using that here, if purely to follow the upstream default while not wasting gratuitous space on disk and the wire - but that's not an obvious win either).

hmaarrfk: I don't understand why it is obvious that it should be opt-in.

To me, nvidia and fat binaries go together. trimming away magma is an optimization.

I think that ultimately we should prefer that for runtime features, users "opt-out" of them, once advanced users have learned their workflows, rather than the surprise of new users being surprised that we are missing a feature.

h-vetinari: > I don't understand why it is obvious that it should be opt-in.

To me the argument is (even though I generally prefer the maximalist approach to packaging!) that 2GB for a feature that few users will touch is a pretty gratuitous waste of bandwith/storage, and those who need it (assuming their percentage among users is reasonably small) should opt-in.

IIUC, you raised a similar [point](https://github.com/conda-forge/ffmpeg-feedstock/pull/279#issuecomment-2526217286) in ffmpeg (about an extra 450MB). That said, I only wrote "supports the argument that it should be opt-in" - I don't feel strongly enough about either option to insist on something.

hmaarrfk: i understsand. it is just not "obvious" thats all.

isuruf: >  few users will touch

Is this based on your experience or facts?

The main driving principle with the pytorch package here has been to provide the same features as in upstream conda/wheels so that users who do use our packages are not driven away by mysterious performance issues. With the deprecation of anaconda.org/pytorch channel, we certainly do not want the users of that channel to look at conda-forge and form the opinion that the pytorch packages in conda-forgeare sub-par. Therefore I do strongly prefer using the same options as the upstream wheels which in this case is with magma.

h-vetinari: > Is this based on your experience or facts?

It's based on @rgommers' statement "10 or so functions, vs. [...] all of libtorch". I don't know pytorch nearly as well as Ralf, so I caveated my point with "assuming their percentage [of those who need it] among users is reasonably small".

IMO it's always a numbers game (and often we don't know the numbers, so we stay conservative; which is fine!) - if we knew, say, that <0.1% of users are affected, then burdening 99.9% users with a useless 2GB download sounds unappealing. The question gets progressively harder the higher we estimate the percentage. Probably anything north of 5% of users, we'd default to performance rather than footprint? I don't have a magic bullet. 🤷 

> provide the same features as in upstream conda/wheels so that users who do use our packages are not driven away by mysterious performance issues.

I agree with the overall direction of course. We still have a few things to catch up (e.g. cpuarch builds, as IIUC pytorch effectively [requires](https://github.com/pytorch/pytorch/blob/v2.5.1/cmake/MiscCheck.cmake#L42-L48) something between x86_64 v3 and v4).

That said, it's certainly technically possible to warn users of a potential performance-cliff where we don't yet support certain upstream optimizations (or require an opt-in for whatever reason). It's mainly a question of how much of that we want to patch and/or upstream.

To be clear, I'm also OK with "fat binaries are preferable to performance degradation or extra maintenance effort". The magma-case is just not so black-and-white IMO. What do you think about the idea of making an exception to link libmagma statically?

rgommers: To be fair, they're ~10 pretty important linear algebra functions. The default matching upstream is also a good argument. I actually wish that was done more often, so I can't complain about the argument being applied here:) For example, cuDNN version is very performance-sensitive, so I wish that were matched better rather than just taking whatever the latest version is.

Static linking or some other way to reduce `libmagma` in size would be good to investigate then though, because binary size is also a performance (and usability) issue.

mgorny:
Replaced the tests with a hand-picked subset. Pulled in a few fixes for numpy-2 testing.

My next idea on the list is to try making it work with rattler-build.

recipe/README.md
isuruf: They are pretty important linear algebra functions like lu_solve, so I would say vast majority of users would use magma. Not the other way. For eg: a simple script like below
```
import torch
a = torch.randn(5, 5).cuda()
b = torch.randn(5, 5).cuda()
torch.linalg.solve(a, b)
```
would invoke magma. The number of functions provided is a poor indicator of its usage. (but certainly valid about code size)

> e.g. cpuarch builds, as IIUC pytorch effectively [requires](https://github.com/pytorch/pytorch/blob/v2.5.1/cmake/MiscCheck.cmake#L42-L48) something between x86_64 v3 and v4).

pytorch requires those from the compiler, but not at runtime. fbgemm does runtime cpu dispatching just like openblas, numpy, etc.

> What do you think about the idea of making an exception to link libmagma statically?

Why do you think that linking libmagma statically will help? Just because upstream used a static build? One option that's different from upstream is https://github.com/conda-forge/libmagma-feedstock/pull/22

isuruf:
Do you have the `libtorch_cuda_linalg.so` artifact without magma somewhere that I can use to check things?

mgorny:
> Do you have the `libtorch_cuda_linalg.so` artifact without magma somewhere that I can use to check things?

Not offhand, but I should have a warm ccache, so I'll get one quickly. However, I've only built for 7.5 CUDA target, is that okay for you?

isuruf:
> is that okay for you?

yep

mgorny:
> Do you have the `libtorch_cuda_linalg.so` artifact without magma somewhere that I can use to check things?

Here's both "magma" and "nomagma" variants (they're small): [linalgs.zip](https://github.com/user-attachments/files/18112090/linalgs.zip)

mgorny:
Ok, some important stuff in today's update:

- I've noticed that I've failed to add `mkl`/`generic` to build string for CUDA + generic builds, and some of them were still skipped — with this version, all 4 variants are built correctly now
- I've also switched from using disjoint build numbers to prioritize CUDA and mkl, to using `track_features`

That said, I think the latter means that I should set the build number to be higher than the highest current build number used, is that correct?

recipe/meta.yaml
```diff
@@ -314,18 +488,18 @@ outputs:
...
+      string: cuda{{ cuda_compiler_version | replace('.', '') }}_{{ blas_impl }}_h{{ PKG_HASH }}_{{ PKG_BUILDNUM }}                  # [megabuild and cuda_compiler_version != "None"]
+      string: cpu_{{ blas_impl }}_h{{ PKG_HASH }}_{{ PKG_BUILDNUM }}                                                                # [megabuild and cuda_compiler_version == "None"]
+      string: cuda{{ cuda_compiler_version | replace('.', '') }}_{{ blas_impl }}py{{ CONDA_PY }}_h{{ PKG_HASH }}_{{ PKG_BUILDNUM }}  # [not megabuild and cuda_compiler_version != "None"]
+      string: cpu_{{ blas_impl }}_py{{ CONDA_PY }}_h{{ PKG_HASH }}_{{ PKG_BUILDNUM }}                                                # [not megabuild and cuda_compiler_version == "None"]
       detect_binary_files_with_prefix: false
-      skip: true  # [cuda_compiler_version != "None" and linux64 and blas_impl != "mkl"]
       # weigh down cpu implementation and give cuda preference
       track_features:
-        - pytorch-cpu                                      # [cuda_compiler_version == "None"]
+        - cpu                                 # [cuda_compiler_version == "None"]
```
isuruf, line 498: It's better to keep the `pytorch-` prefix here

recipe/meta.yaml
```diff
@@ -1,16 +1,15 @@
 {% set version = "2.5.1" %}
 {% set build = 6 %}
 
-{% if cuda_compiler_version != "None" %}
-{% set build = build + 200 %}
-{% endif %}
```
isuruf, line 6: Let's keep these.

mgorny: I'm not sure if we can keep them while switching to rattler-build. I've removed these conditional macros to make conversion easier.

isuruf: I don't see a reason why we can't.

isuruf:
It seems like with just `libtorch_cuda_linalg.so` replaced with `nomagma` variant, you still have `torch.cuda.has_magma == True`

recipe/build_common.sh
```diff
@@ -0,0 +1,268 @@
...
+fi
+
+echo '${CXX}'=${CXX}
+echo '${PREFIX}'=${PREFIX}
+
+case ${PKG_NAME} in
+  libtorch-split)
+    # Call setup.py directly to avoid spending time on unnecessarily
+    # packing and unpacking the wheel.
+    $PREFIX/bin/python setup.py build
```
isuruf, line 207: Calling `setup.py` directly is deprecated right?

mgorny: Depends on who you ask. setuptools itself only deprecated `setup.py install` so far. Then, the hacks PyTorch do on top of setuptools are all pretty much incompatible with PEP517 (especially `sys.argv` manipulations).

mgorny:
I've reverted the build number and `track_features` changes.

mgorny:
> It seems like with just `libtorch_cuda_linalg.so` replaced with `nomagma` variant, you still have `torch.cuda.has_magma == True`

Hmm, I suppose there's no trivial way around that. We could then multiple all `pytorch` variants for that but that sounds like an overkill.

mgorny:
Patched `hasMAGMA()` to detect the used variant by presence of `libmagma.so`, as suggested by @isuruf. While at it, fixed the `run_constrained` to actually uninstall libmagma.

recipe/patches/0013-Adjust-CUDAHooks-hasMAGMA-for-libtorch_cuda_linalg.s.patch
```diff
@@ -0,0 +1,53 @@
...
++  // installed.
++  Dl_info info;
++  if (dladdr(CUDAHooks::hasMAGMA, &info)) {
++    std::string path = info.dli_fname;
++    char* directory = dirname(path.data());
++    path = directory;
++    path += "/libmagma.so";
++    printf("%s\n", path.c_str());
++  }
++#endif
```
isuruf, line 46: This doesn't return anything right?

mgorny: Ah, sorry, forgot to `git commit` the non-debug version of the patch :facepalm:.

mgorny: Fixed now.

h-vetinari:
FWIW, osx+generic has some failures for 32-bit complex types, where seemingly the entire imaginary part of the result is dropped. On the one hand that's pretty bad, on the other hand it's a niche type for a setup (non-mkl) that's not really supported upstream. I'm kind of tempted to just skip those for now. Any thoughts?
```
=========================== short test summary info ============================
FAILED [0.1311s] test/test_linalg.py::TestLinalgCPU::test_matmul_small_brute_force_1d_Nd_cpu_complex64 - AssertionError: Scalars are not close!

Expected (32.411563873291016+27.98528289794922j) but got 0j.
Absolute difference: 42.82155452094975 (up to 5e-05 allowed)
Relative difference: 1.0 (up to 0.0001 allowed)
torch.Size([1]) x torch.Size([1]) = torch.Size([])

    python test/test_linalg.py TestLinalgCPU.test_matmul_small_brute_force_1d_Nd_cpu_complex64

FAILED [0.0124s] test/test_linalg.py::TestLinalgCPU::test_1_sized_with_0_strided_cpu_float32 - AssertionError: Tensor-likes are not close!

Mismatched elements: 1 / 4096 (0.0%)
Greatest absolute difference: 0.00010824203491210938 at index (3, 0, 495) (up to 0.0001 allowed)
Greatest relative difference: 0.00011769433331210166 at index (3, 0, 495) (up to 1.3e-06 allowed)

    python test/test_linalg.py TestLinalgCPU.test_1_sized_with_0_strided_cpu_float32

FAILED [0.0071s] test/test_linalg.py::TestLinalgCPU::test_dot_vs_numpy_cpu_complex64 - AssertionError: Scalars are not close!

Expected (-0.14432667195796967-0.14127731323242188j) but got 0j.
Absolute difference: 0.20196402519417966 (up to 0.0001 allowed)
Relative difference: 1.0 (up to 1.3e-06 allowed)

    python test/test_linalg.py TestLinalgCPU.test_dot_vs_numpy_cpu_complex64

FAILED [0.0078s] test/test_linalg.py::TestLinalgCPU::test_vdot_vs_numpy_cpu_complex64 - AssertionError: Scalars are not close!

Expected (-0.2586096227169037-0.33168256282806396j) but got 0j.
Absolute difference: 0.42058561488235885 (up to 0.0001 allowed)
Relative difference: 1.0 (up to 1.3e-06 allowed)

    python test/test_linalg.py TestLinalgCPU.test_vdot_vs_numpy_cpu_complex64

= 4 failed, 7562 passed, 1385 skipped, 31 xfailed, 74107 warnings in 1198.05s (0:19:58) =
```

rgommers:
> FWIW, osx+generic has some failures for 32-bit complex types, where seemingly the entire imaginary part of the result is dropped. On the one hand that's pretty bad, on the other hand it's a niche type for a setup (non-mkl) that's not really supported upstream. I'm kind of tempted to just skip those for now. Any thoughts?

That is probably gh-180, which is a long-standard issue. Agreed it's pretty bad, but seems best not to couple to this PR so skipping seems reasonable.

h-vetinari:
For now I'm waiting the outcome of the linux-64 MKL job, resp the linux-aarch one to see under what conditions this triggers. Of course, if someone knows what's happening (@isuruf @mgorny), elucidating comments are most welcome.

h-vetinari:
> I think we might need to make [f51899c](https://github.com/conda-forge/pytorch-cpu-feedstock/commit/f51899cc3e8298d9d3fd677df47f822e1cf08556) conditional. It's necessary on osx-64 [...]

nevermind, osx + generic still fails as before

h-vetinari:
So we have failures for `linux-64` and `osx-64` with `generic` BLAS, whereas MKL passes on both of those platforms. Moreover, the change from #304 is accurately titled, in that the upstream [logic](https://github.com/pytorch/pytorch/blob/v2.5.1/cmake/Modules/FindBLAS.cmake#L320-L384) for these variables never triggers, as it's guarded by `IF (CMAKE_CROSSCOMPILING)`.

In #299, I'm trying to see what happens if we apply the fix from #304 for all `generic` builds, regardless of whether it's native or cross-compiled. If that works, then that would hopefully (🤞) be the last required update to get this PR green & merged. If it doesn't work, I suggest we skip the tests for now (@hmaarrfk asked to uncouple these concerns in #304).

hmaarrfk:
The one thing I find missing is a single sentence on the top comment that expresses the agreed intention of what someone should has Cuda installed should experience:

is magma opt in or opt out?

h-vetinari:
> is magma opt in or opt out?

See the [thread](https://github.com/conda-forge/pytorch-cpu-feedstock/pull/298#discussion_r1874819009) further up. It's opt-out, to match upstream.

recipe/README.md
h-vetinari: Somehow missed the last ping here. Thanks for the info, you certainly know pytorch much better than I do.

> Why do you think that linking libmagma statically will help?

Presumably those 10 pytorch functions do not make use of the entirety of the 2GB libmagma. If they were compiled against a static build, the impact would be much smaller. 

> One option that's different from upstream is [conda-forge/libmagma-feedstock#22](https://github.com/conda-forge/libmagma-feedstock/pull/22)

That's even better though! :)

jakirkham: Including this bit from that merged PR thread:

Me:
> Did you measure the difference between the two approaches? Would be interested to know how they compare

Isuru:
> 2GB -> 620 MB

xref: https://github.com/conda-forge/libmagma-feedstock/pull/22#issuecomment-2533448783

<hr>

IOW a nearly 70% reduction in size. Perhaps it is worth rechecking the analysis given this change? The biggest thing might be something else now

mgorny:
Hmm, I see some of the jobs timed out in the end. Should I try to reduce the number of tests run?

mgorny:
> So we have failures for `linux-64` and `osx-64` with `generic` BLAS, whereas MKL passes on both of those platforms. Moreover, the change from #304 is accurately titled, in that the upstream [logic](https://github.com/pytorch/pytorch/blob/v2.5.1/cmake/Modules/FindBLAS.cmake#L320-L384) for these variables never triggers, as it's guarded by `IF (CMAKE_CROSSCOMPILING)`.

Hmm, this is quite interesting, because I'm quite sure these tests passed for me locally. I'll retry again later today.

h-vetinari:
> Hmm, I see some of the jobs timed out in the end. Should I try to reduce the number of tests run?

No, I cancelled runs I knew would fail to conserve resources. I did restart the non-CUDA aarch job because I wanted to see if it worked there.

h-vetinari:
> Hmm, this is quite interesting, because I'm quite sure these tests passed for me locally. I'll retry again later today.

Isuru had an update in #304 that I've already picked into this branch. I haven't pushed it yet because I'm waiting for the outcome of the aarch CI.

mgorny:
> Hmm, this is quite interesting, because I'm quite sure these tests passed for me locally. I'll retry again later today.

Hmm, I can reproduce. No clue why I missed it.

h-vetinari:
Perhaps you always built the MKL variant? 🤔 

In any case, the aarch builds here seem to be stuck. It should take 3-4h for that run, and we're already at 7h now. Not sure if https://github.com/conda-forge/pytorch-cpu-feedstock/pull/298/commits/434b9e5bf1f68ccb38703d203426c80a6a7d219b tripped something up (at the time I was looking for something non-banal to commit in order to retrigger a CI run); I can rebase that out of course if we prefer.

mgorny:
Might be that `pytest` hangs somewhere, and the build job didn't flush the output. My personal preference would be to merge what we have ASAP, and work on further changes on top of that.

h-vetinari:
OK, the osx-64 + generic build in #299 passed the complex64 tests (there one last failing test that's just a tolerance violation which we can skip); I'm going to push this here, but remove the aarch testing changes. This PR is big enough to not open up another issue.

mgorny:
Just to be clear, this version does not feature an explicit `wheel` dependency. The optimizations removed packing and unpacking the wheel, and the only remaining `wheel` use is the vendored dependency in setuptools.

h-vetinari:
> Just to be clear, this version does not feature an explicit `wheel` dependency. The optimizations removed packing and unpacking the wheel, and the only remaining `wheel` use is the vendored dependency in setuptools.

Ah great, I had picked that up from #304; wasn't aware this got removed!

recipe/meta.yaml
```diff
@@ -125,54 +149,217 @@ requirements:
...
+      skip: true  # [win or cuda_compiler_version == "None"]
+      ignore_run_exports_from:
+        - python *                               # [megabuild]
+        - numpy *                                # [megabuild]
+        - cross-python_{{ target_platform }}     # [megabuild and build_platform != target_platform]
+      ignore_run_exports:
+        - python *                               # [megabuild]
+        - numpy *                                # [megabuild]
+      track_features:
+        - nomagma                                # [not use_magma]
```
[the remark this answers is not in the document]
mgorny: That won't work since both variants are built simultaneously.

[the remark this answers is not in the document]
mgorny: Well, or at least I think it won't, I may be confused by how templates are rendered.

mgorny:
Also, I've just noticed a typo in `pytorch-gpu` run requirements. I'll push a fix after quick testing.

recipe/meta.yaml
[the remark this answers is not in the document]
isuruf: Doing anything at the top level using the variable `use_magma` will result in the job being split into two. I pushed a commit that should work.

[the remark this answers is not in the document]
mgorny: Just confirmed that @isuruf's commit works.

mgorny:
> I'm happy to edit the top level comment, but some people see it as too aggressive.

For the record, I don't mind that at all.

recipe/meta.yaml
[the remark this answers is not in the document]
hmaarrfk: awesome thank you both!

h-vetinari:
@mgorny, given how frequently you contribute on this feedstock now, could you add yourself to the [access list](https://github.com/Quansight/open-gpu-server/#access) of the open-gpu server? That way, when you push changes, it won't require another push by someone else to trigger a run.

(there's a technical limitation that only commits from users on the access list can be run at all; it's not possible to "approve" a CI run for a commit from a user not on the list)

mgorny:
> @mgorny, given how frequently you contribute on this feedstock now, could you add yourself to the [access list](https://github.com/Quansight/open-gpu-server/#access) of the open-gpu server? That way, when you push changes, it won't require another push by someone else to trigger a run.
> 
> (there's a technical limitation that only commits from users on the access list can be run at all; it's not possible to "approve" a CI run for a commit from a user not on the list)

Requested, thanks!

h-vetinari:
The MKL+CUDA job ran for an  unusually long time (just over 12h), but didn't actually fail
```
linux= 7568 passed, 1383 skipped, 31 xfailed, 75823 warnings in 1621.53s (0:27:01) ==
linux+ pip check
linuxNo broken requirements found.
linux+ python -c 'import torch; print(torch.__version__)'
linux2.5.1.post207
linux+ python -c 'import torch; torch.tensor(1).to('\''cpu'\'').numpy(); print('\''numpy support enabled!!!'\'')'
linuxnumpy support enabled!!!
linux+ test -f $PREFIX/lib/libtorch_python.so
linux+ exit 0
linux##[error]The operation was canceled.
```

Perhaps this was just an unlucky run w.r.t. resource congestion (normally CUDA runs finish in 8-9h). In any case, the run had basically reached the very end (passed the tests), so we may want to raise
https://github.com/conda-forge/pytorch-cpu-feedstock/blob/b498f465ef7ecd80c53041f914cd82dca85a2c91/conda-forge.yml#L16
a bit, e.g. to 900 (=15h)

recipe/README.md
rgommers: It looks even better on my machine after @isuruf's compression fix:

```
$ mamba create -n pytorch-gpu-19dec2024 pytorch-gpu
$ cd ~/mambaforge/envs/pytorch-gpu-19dec2024/
$ du -hsx * | sort -rh | head -3
4,1G    lib
1,6G    targets
69M     include

$ du -hsx lib/* | sort -rh | head -10
1,4G    lib/libtorch_cuda.so
432M    lib/libcudnn_engines_precompiled.so.9.3.0
296M    lib/python3.13
293M    lib/libmagma.so
276M    lib/libtorch_cpu.so
252M    lib/libnccl.so.2.23.4
233M    lib/libcudnn_adv.so.9.3.0
104M    lib/libcudnn_ops.so.9.3.0
91M     lib/libmagma_sparse.so
68M     lib/libmkl_core.so.2

$ du -hsx targets/x86_64-linux/lib/* | sort -rh | head -10
469M    targets/x86_64-linux/lib/libcublasLt.so.12.6.4.1
280M    targets/x86_64-linux/lib/libcusparse.so.12.5.4.2
266M    targets/x86_64-linux/lib/libcufft.so.11.3.0.4
146M    targets/x86_64-linux/lib/libcusolver.so.11.7.1.2
104M    targets/x86_64-linux/lib/libcublas.so.12.6.4.1
92M     targets/x86_64-linux/lib/libcurand.so.10.3.7.77
86M     targets/x86_64-linux/lib/libcusolverMg.so.11.7.1.2
56M     targets/x86_64-linux/lib/libnvrtc.so.12.6.85
49M     targets/x86_64-linux/lib/libnvJitLink.so.12.6.85
5,1M    targets/x86_64-linux/lib/libnvrtc-builtins.so.12.6.85
```

So 385 MB left for `libmagma.so` + `libmagma_sparse.so`.

I'll also note that there are two `linux-64/libmagma` [packages](https://anaconda.org/conda-forge/libmagma/files), and the CUDA 12 one seems to be a lot smaller than the CUDA 11 one. There are no interpretable build strings, but comparing package hashes with the logs from https://github.com/conda-forge/libmagma-feedstock/pull/22 shows that CUDA 11 is the larger one. The analysis above uses the `h7847c38_1` package.

image

mgorny: Well, I'm not opposed to reverting the magma parts. Just let me know if I should do it.

h-vetinari (approved):
Thanks a lot for the work on this! I think we should merge this as-is, and restart after merge if necessary. The next PR can increase the timeout.

h-vetinari:
Well of course... Just after I commented, the aarch CPU job failed (despite the aarch CUDA job passing, which I had taken as a sign that things are looking good, not remembering that tests are skipped for CUDA builds):
```
=========================== short test summary info ============================
FAILED [1192.7443s] test/test_linalg.py::TestLinalgCPU::test_pca_lowrank_cpu - torch._C._LinAlgError: linalg.svd: The algorithm failed to converge because the input matrix contained non-finite values.

    python test/test_linalg.py TestLinalgCPU.test_pca_lowrank_cpu

FAILED [20.3430s] test/test_autograd.py::TestAutograd::test_pynode_destruction_deadlock - AssertionError: Example code timed out! See the code sample in the test for details.

    python test/test_autograd.py TestAutograd.test_pynode_destruction_deadlock

FAILED [1.1653s] test/test_linalg.py::TestLinalgCPU::test_cholesky_cpu_float32 - torch._C._LinAlgError: linalg.cholesky: (Batch element 0): The factorization could not be completed because the input is not positive-definite (the leading minor of order 94 is not positive-definite).

    python test/test_linalg.py TestLinalgCPU.test_cholesky_cpu_float32

FAILED [4591.8560s] test/test_linalg.py::TestLinalgCPU::test_svd_lowrank_cpu_complex128 - torch._C._LinAlgError: linalg.svd: The algorithm failed to converge because the input matrix contained non-finite values.

    python test/test_linalg.py TestLinalgCPU.test_svd_lowrank_cpu_complex128

FAILED [1286.1502s] test/test_linalg.py::TestLinalgCPU::test_svd_lowrank_cpu_float64 - torch._C._LinAlgError: linalg.svd: The algorithm failed to converge because the input matrix contained non-finite values.

    python test/test_linalg.py TestLinalgCPU.test_svd_lowrank_cpu_float64

FAILED [2867.8001s] test/test_mkldnn.py::TestMkldnnCPU::test_lstm_cpu - RuntimeError: could not create a primitive descriptor for an LSTM backward propagation primitive

    python test/test_mkldnn.py TestMkldnnCPU.test_lstm_cpu

= 6 failed, 7582 passed, 1363 skipped, 31 xfailed, 75802 warnings in 22702.77s (6:18:22) =
```

One thing to note is the extreme runtime in emulation - I think we'll want to further reduce testing on aarch. The extreme runtime is probably also responsible for at least some of the failures (at least the timeout, probably also the non-convergence ones). The one failure that doesn't fall under this is `test_cholesky_cpu_float32` -- I'm not surprised that the failure would happen with non-default floating point types (as those are comparatively less well-tested in all components involved: compilers, openblas, QEMU, pytorch, etc.).

Any suggestions how to get the testing runtime in emulation under a reasonable limit (say 1h) @mgorny? In any case, I think we can skip the failing tests here for now (or even all of the test suite run; we didn't have that before this PR at all).

isuruf:
With 385MB of savings, is this worth the additional maintenance burden?

Tobias-Fischer:
> With 385MB of savings, is this worth the additional maintenance burden?

I was wondering the same - especially considering that I guess only a very small minority will opt-out.

h-vetinari:
@isuruf, I presume you're responding to
> @rgommers: So 385 MB left for `libmagma.so` + `libmagma_sparse.so`.

> @mgorny; Well, I'm not opposed to reverting the magma parts. Just let me know if I should do it.

I'm in favour of keeping this PR as is. The <400MB by default are less than ~10% of the pytorch footprint, and people who really need the space can use the nomagma variant.

isuruf:
> people who really need the space can use the nomagma variant.

Who are they? If nobody speaks up with a real use-case, I'd rather we close this PR and start a new one with the tests and the other goodies in this PR.

h-vetinari:
> Who are they? If nobody speaks up with a real use-case

This is addressing #275 and a couple people have mentioned concerns about the size of the footprint. 🤷 

I'm not sure I understand where you're coming from here - @mgorny said in the OP:
> @isuruf helped me a lot with this, particularly with refactoring the builds, so both variants are built in one run.

So I would have thought you're on board with the idea, resp. the details we hashed out (magma as a shared library, enabled by default, with an opt-out). I'm not proposing to change this.

If you're now saying "wait, why are we making magma optional at all?" then I'll have to leave the floor to the people who were concerned about the footprint (e.g. @zklaus @rgommers IIUC) - I don't have strong feelings about this (well, except perhaps about getting this PR over the finish line 😅 but if people want to remove the magma-changes again, fine by me)

h-vetinari:
FWIW, even if we make magma unconditional again, it'd likely be less work to merge this PR as-is and then remove the magma opt-out with the next PR again, rather than rework the history of this PR to never make magma optional in the first place.

mgorny:
I also think it'd be cleaner to remove it on top of the existing batch, if only to preserve the history and discussion here. Of course, I'm open to the other way as well, if that's what you prefer. For the former variant, I can start preparing the commits for the next PR pretty much immediately.

mgorny:
One more thought: if we're going to unsplit linalg, I suppose it makes sense not to publish the split package in the first place. I've created https://github.com/mgorny/pytorch-cpu-feedstock/pull/2 with my current changes that is based on this PR's branch — if you think we should do it this way, I can merge it to update this PR.

recipe/README.md
rgommers: > Well, I'm not opposed to reverting the magma parts. Just let me know if I should do it.

For completeness here: Isuru moved this decision conversation to gh-275, which seems like the right thing to do for visibility.

h-vetinari:
Sigh, it was quite non-trivial to get the rerender not to build the magma jobs, but it seems I still messed something up in https://github.com/conda-forge/pytorch-cpu-feedstock/pull/298/commits/62bb757fcd549f01524c7484bdc435f9b4eb8260 (and because the `set -ex` only comes afterwards, I don't even have logs for what happened). Likewise the syntax in https://github.com/conda-forge/pytorch-cpu-feedstock/pull/298/commits/5c845497cb9a7916c8ae07dcb410a43e72a08ce0 didn't work (though the `bool * list` pattern has worked [fine](https://github.com/conda-forge/faiss-split-feedstock/blob/f310252930523c239e64a37db2d9dc61d727dcaa/recipe/meta.yaml#L30) in other such places).

If only from the side of handling conda-smithy, I'm in favour or removing the magma opt-out again.

h-vetinari:
@aktech, it seems we're back to failing GPUs on the open-gpu-server - right off the bat 2 jobs fail with:
```
  docker: Error response from daemon: failed to create task for container: failed to create shim task: OCI runtime create failed: runc create failed: unable to start container process: error during container init: error running hook #0: error running hook: exit status 1, stdout: , stderr: Auto-detected mode as 'legacy'
  nvidia-container-cli: initialization error: nvml error: driver not loaded: unknown.
  time="2024-12-20T21:45:54Z" level=error msg="error waiting for container: "
```

I cancelled & restarted the workflow to double-check, and the exact same thing happened again.

aktech:
Hey @h-vetinari apologies for the inconvenience. That's expected, I was trying to debug the hardware problem with the problematic GPUs. I'll revert now, will chose another time for this.

h-vetinari:
Ah, that makes sense of course!

I don't want to block  your work of course - I'm sure we can find a way to coordinate a time? From my POV: after this PR is merged and the builds published would probably be a good time (though not sure if you'll be working over the holidays) - in any case, I think it would already be good enough if you simply let us know when you take things offline, if you see active ongoing PRs here.



> I'll revert now, will chose another time for this.

Can you let us know when things should be operational ready again? I restarted ~15min after your comment, but there's still a failure. I'll cancel the worfklow for now.

aktech:
Sure, let's get this one in, then I can do the debugging over a quieter time, we can coordinate time via the Zulip chat.

>Can you let us know when things should be operational ready again? I restarted ~15min after your comment, but there's still a failure. I'll cancel the worfklow for now.

I missed removing one of the faulty one, it should work now.

recipe/build.sh
```diff
@@ -1,229 +1,3 @@
...
-  cp build/CMakeCache.txt build/CMakeCache.txt.orig
-else
-  # Keep this in ${PREFIX}/lib so that the library can be found by
-  # TorchConfig.cmake.
-  # With upstream non-split build, `libtorch_python.so`
-  # and TorchConfig.cmake are both in ${SP_DIR}/torch/lib and therefore
-  # this is not needed.
-  mv ${SP_DIR}/torch/lib/libtorch_python${SHLIB_EXT} ${PREFIX}/lib
-fi
+# we are using a separate file here to avoid conda-build thinking that
```
Tobias-Fischer, line 1: Is a separate build_common.sh still necessary if we don’t end up having non-magma builds?

recipe/meta.yaml
```diff
@@ -9,47 +9,44 @@
 {% set build = build + 100 %}
 {% endif %}
 
+{% set mkl = "<2024" %}
+
 package:
-  name: libtorch
+  name: libtorch-split
```
isuruf, line 15: Let's not do this unless absolutely necessary. I was okay doing this because of libmagma builds, but now that they are gone, let's not do this split build and keep the top level as `libtorch`.

mgorny: Do you prefer me reverting it as more commits, or rebasing the whole thing?

isuruf: Whichever is easier for you

mgorny: Ok, I went for more commits, because I wanted to retain full history and attribution, and collapsing the code back seems relatively clean anyway. I'm going to finish testing in a few minutes and push.

recipe/meta.yaml
```diff
@@ -291,41 +286,74 @@ outputs:
...
         # At conda-forge, we target versions of OSX that are too old for MPS support
         # But if users install a newer version of OSX, they will have MPS support
         # https://github.com/conda-forge/pytorch-cpu-feedstock/pull/123#issuecomment-1186355073
         # - python -c "import torch; assert torch.backends.mps.is_available()" # [osx]
+
+        # python-version-specific library (default location in SP_DIR symlinks back to this)
         - test -f $PREFIX/lib/libtorch_python${SHLIB_EXT}     # [unix]
 
+        # a reasonably safe subset of tests that should run under 15 minutes
+        # disable hypothesis because it randomly yields health check errors
```
danpetry, line 304: Have you considered their [smoke test](https://github.com/pytorch/builder/blob/main/test/smoke_test/smoke_test.py)?

h-vetinari: AFAIU, the testing here is more comprehensive than the smoke test. But that shouldn't stop us from adding the smoke test in the next PR if it's considered useful (and potentially covers some aspects not covered by the choice of modules here)

mgorny:
Hmm, that's a lot of test failures for the `linux_64_blas_implgenericc_compiler_version13cuda_compilercuda-nvcccuda_compiler_version12.6cxx_compiler_version13` job. Here, I'm seeing 5 now:

```
FAILED [0.0019s] test/test_custom_ops.py::TestCustomOpAPI::test_compile - Run...
FAILED [0.0049s] test/test_custom_ops.py::TestCustomOpAPI::test_fake - Runtim...
FAILED [0.0018s] test/test_custom_ops.py::TestCustomOp::test_data_dependent_compile
FAILED [0.0018s] test/test_custom_ops.py::TestCustomOp::test_functionalize_error
FAILED [0.0134s] test/test_torch.py::TestTorch::test_print - AssertionError: ...
```

But I don't have a GPU here.

recipe/meta.yaml
```diff
@@ -291,41 +286,74 @@ outputs:
...
+        {% set skips = skips ~ " or test_pynode_destruction_deadlock" %}                # [aarch64]
+        {% set skips = skips ~ " or (TestLinalgCPU and test_cholesky_cpu_float32)" %}   # [aarch64]
+        {% set skips = skips ~ " or (TestLinalgCPU and test_pca_lowrank_cpu)" %}        # [aarch64]
+        {% set skips = skips ~ " or (TestLinalgCPU and test_svd_lowrank_cpu)" %}        # [aarch64]
+        {% set skips = skips ~ " or (TestMkldnnCPU and test_lstm_cpu)" %}               # [aarch64]
+        # dynamo does not support python 3.13
+        {% set skips = skips ~ " or (TestCustomOp and test_data_dependent_compile)" %}  # [py==313]
+        {% set skips = skips ~ " or (TestCustomOp and test_functionalize_error)" %}     # [py==313]
+        {% set skips = skips ~ " or (TestCustomOpAPI and test_compile)" %}              # [py==313]
+        {% set skips = skips ~ " or (TestCustomOpAPI and test_fake)" %}                 # [py==313]
```
h-vetinari, line 331: @mgorny, it seems that the failures match the tests here? So then it's just the selector that needs to be expanded

h-vetinari: Although it's strange - the skip for `TestTorch::test_print` is unconditional; I'm on a phone right now so cannot debug further, but very weird how the skips don't take effect.

recipe/README.md
```diff
@@ -98,3 +101,77 @@ of 2024-11-28:
...
+
+1. For the top-level rule (`libtorch-split`), we perform the base environment
+   setup and run `setup.py build` to build the libraries and collect the data
+   files without actually installing them. Then we move the files we need
+   into temporary directories for repackaging.
+
+   a. If `megabuild` is enabled, we build against a fixed Python version.
+      Otherwise, we build using the final Python version.
+
+2. For the `libtorch` package, we manually install files that were prepared earlier.
```
isuruf, line 172: Let's update these 2 bullet points.

isuruf (approved):

recipe/meta.yaml
mgorny: At a first glance, it seems that these are passed to pytest correctly, so my guess would be that `::` syntax doesn't work there. I'm going to try experimenting with it a little more in a while, lemme just tidy up some other stuff first.

mgorny: Confirmed that neither `::` nor `.` works — I guess you can't match both class name and method name via `-k`. Good news is that with the subset of tests we're running, it should be fine to just match on name, so I'll do that. I'll also do that for aarch64 skips, since I suspect they didn't really work as well.

mgorny: Oh, sorry, I've just realized I didn't pull the newest changes yet.

recipe/README.md
mgorny: Done. I'll push once I've dealt with the tests as well.

recipe/meta.yaml
mgorny: Nevermind these, after pulling all tests pass for me. Perhaps https://github.com/conda-forge/pytorch-cpu-feedstock/actions/runs/12448456323/job/34771658661?pr=298 was a fluke?

mgorny:
To be honest, I'm a bit worried about how much we're spending on failing tests here. Perhaps we could cancel all test runs, except for the one that has failed before?

I'm also wondering if we shouldn't just either ignore failures (i.e. restore `|| :`) or skip them for now, especially if they fail again, merge this as-is (as we know it's not a regression), and get them to work separately.

mgorny:
Just ran the complete build for `linux_64_blas_implgenericc_compiler_version13cuda_compilercuda-nvcccuda_compiler_version12.6cxx_compiler_version13` (i.e. the one that failed previously) on our sm75-enabled server and all tests passed.

I've also prepared one more change requested by @isuruf in https://github.com/mgorny/pytorch-cpu-feedstock/pull/3 ; not merging it yet, since I don't want to restart CI.

recipe/meta.yaml
danpetry: Side note, if we're bringing across changes from Anaconda's recipe to here, would you prefer to have each change in a separate PR or all in one PR?

h-vetinari:
> To be honest, I'm a bit worried about how much we're spending on failing tests here.

Yeah, let's just skip them on GPUs. Especially now that it turns out that the x64 + CUDA build can fail catastrophically:
```
sssssssssssssssssssssssssssssssssssssssssssssssssssssssF.....ssss.F.ssss [ 40%]
......ssssss..ss............ssssssssssss..............F.....F....F.....F [ 40%]
..FFFFF.FFFFFsFFFFFFF....F.FFF..F.F...F.F..s.F.sFF.............FFFFF.sss [ 41%]
s.F.FFFFFFFFFF...FxxFFFFFFFFFFFFFFFFF..FFFFF.FsFsssFxxF..FFFF........... [ 41%]
..FFFFFFFFFFF........................F..F.Fs..Fs.............FFFFFFFFFFF [ 42%]
..FFFFFFxxFFFFFFFFFFF.FFFFF.F..FFF.....F..........F......F...FFFFF.F...F [ 42%]
.FF.......FFFFFFFFF............................................FFFFFFFFF [ 43%]
FFFF............................FFF..FF................................. [ 43%]
......F............FF.F.................................FFFFFFFF.s...... [ 43%]
.............F..........................................FFFF.FF.FFFF.... [ 44%]
...........xFxxx...F..F.......F.....F....F.s....F............FFxxFxxF... [ 44%]
.F........F.........F....F................F................F............ [ 45%]
.......F...............................................................F [ 45%]
[...]
..F.............F.F...F..s.F.F....s.sF...F..sF.FF.FF..F.F............... [ 58%]
./.scripts/run_docker_build.sh: line 108:  2497 Killed                  docker run ${DOCKER_RUN_ARGS} -v "${RECIPE_ROOT}":/home/conda/recipe_root:rw,z,delegated -v "${FEEDSTOCK_ROOT}":/home/conda/feedstock_root:rw,z,delegated -e CONFIG -e HOST_USER_ID -e UPLOAD_PACKAGES -e IS_PR_BUILD -e GIT_BRANCH -e UPLOAD_ON_BRANCH -e CI -e FEEDSTOCK_NAME -e CPU_COUNT -e BUILD_WITH_CONDA_DEBUG -e BUILD_OUTPUT_ID -e flow_run_id -e remote_url -e sha -e BINSTAR_TOKEN -e FEEDSTOCK_TOKEN -e STAGING_BINSTAR_TOKEN "${DOCKER_IMAGE}" bash "/home/conda/feedstock_root/${PROVIDER_DIR}/build_steps.sh"
##[error]Process completed with exit code 137.
```

h-vetinari:
> Especially now that it turns out that the x64 + CUDA build can fail catastrophically:

On second thought @mgorny, this might actually be due to one of the following commits of yours, which were part of https://github.com/mgorny/pytorch-cpu-feedstock/pull/2, but mostly tangential to this PR:
* cbe2c420d4a8a39aa87790e72e8400f26e51425a
* 1b15f6a777de9c8b58a30c0bd914c67e6d9c455a
* 27c5e9913297e2e64d3252e95a81e15f5ade6fdf

AFAICT, we haven't had a x64+generic+CUDA job since 98f8a9c3b5a65224b33c80a72fbee18fe71ad7dc that ran and passed the test suite (modulo the misapplied skips). Both times that the python test suite ran on that job (97cb097fad092e31861923c3f70afd3d9b5d49e4 & 4fe6a1edc19102748005ba486d7ca48531087ae3), things failed catastrophically.

I'm going to roll back those commits (& https://github.com/conda-forge/pytorch-cpu-feedstock/pull/298/commits/0a76034e457248c17847e18c88499095ff80e4fe) for now; we can do more focussed debugging on the enablement in a separate PR.

hmaarrfk:
If these are green, I would be in strong favor or just merging this as is an cleaning up in separately.

The main concern I have is loosing momentun on nit picks.

If the test suite doesn't pass, we should disable the tests, and work on a separate PR to address the failing tests so as to allow the following other projects to move forward:

* https://github.com/conda-forge/pytorch-cpu-feedstock/pull/304
* https://github.com/conda-forge/pytorch-cpu-feedstock/pull/280
* https://github.com/conda-forge/pytorch-cpu-feedstock/pull/231

(In no order of preference from me)

I've also updated the top level description to be more in line with the new goals of the PR.

mgorny:
The crashing-so-far generic + CUDA job has passed this time. Can we merge it now? I don't think there's a point in waiting for the remaining jobs to run.

Also, should I open a separate PR to redo the extra deps that have caused trouble here? Possibly one by one, to see which one was responsible.

hmaarrfk:
@mgorny did you have any other pending commits you wanted to get in? 

I think we should just wait since the builds have "started" the problems often occur if the builds are "stuck".

mgorny:
No, not at the moment.

All checks passed!

recipe/meta.yaml
h-vetinari: @danpetry: Depends a bit on the quantity of changes, and how well they're separated in terms of commits. But you can start with everything you have in mind, and we can carve out smaller PRs if necessary.

danpetry:
ok, good to know.
Daniel Petry
Sr. Software Engineer
***@***.***
<https://www.anaconda.com/?utm_source=email-signature&utm_medium=anaconda>
www.anaconda.com
<https://www.anaconda.com/?utm_source=email-signature&utm_medium=anaconda>
Sign up
<https://www.anaconda.com/newsletter?utm_source=email-signature&utm_medium=anaconda>
 for our newsletter
<https://www.facebook.com/anacondainc> <https://twitter.com/anacondainc>
<https://www.linkedin.com/company/anacondainc/>
<https://github.com/ContinuumIO>

PR #300, opened 2024-12-06
Allow CUDA path override only during package build

jeongseok-meta:
It seems the modification to `FindCUDAToolkit.cmake` in https://github.com/conda-forge/pytorch-cpu-feedstock/commit/55c97fe71c062d6cf81b6cc554759ee1c03b5b26#r150044376 has caused downstream build failures:
- https://github.com/conda-forge/torchaudio-feedstock/pull/12#discussion_r1870665348
- https://github.com/facebookincubator/momentum/actions/runs/12207717521/job/34059648061#step:8:3777

The key issues are that the modification alters the behavior of finding CUDAToolkit paths and it is unclear whether this change was intended to affect downstream packages or if it was solely for building the package itself.

**Current "Fix"**: One possible solution is to modify the behavior only for building the pytorch package, while leaving the behavior unchanged when it's used by downstream packages. However, this may not be the ideal solution, and we would like to discuss alternative approaches.

Please share your thoughts and suggestions on how to proceed.

Checklist

### DIFF
--- a/recipe/build.sh
+++ b/recipe/build.sh
@@ -3,7 +3,7 @@
 set -ex
 
 # This is used to detect if it's in the process of building pytorch
-export IN_PYTORCH_BUILD
+export IN_PYTORCH_BUILD=1
 
 # https://github.com/conda-forge/pytorch-cpu-feedstock/issues/243
 # https://github.com/pytorch/pytorch/blob/v2.3.1/setup.py#L341
--- a/recipe/build.sh
+++ b/recipe/build.sh
@@ -2,6 +2,7 @@
 
 set -ex
 
+# This is used to detect if it's in the process of building pytorch
 export IN_PYTORCH_BUILD
 
 # https://github.com/conda-forge/pytorch-cpu-feedstock/issues/243
--- a/recipe/build.sh
+++ b/recipe/build.sh
@@ -2,6 +2,8 @@
 
 set -ex
 
+export IN_PYTORCH_BUILD
+
 # https://github.com/conda-forge/pytorch-cpu-feedstock/issues/243
 # https://github.com/pytorch/pytorch/blob/v2.3.1/setup.py#L341
 export PACKAGE_TYPE=conda
--- a/recipe/patches/0010-Allow-overriding-CUDA-related-paths.patch
+++ b/recipe/patches/0010-Allow-overriding-CUDA-related-paths.patch
@@ -17,7 +17,7 @@ index ec9ae530..76d1fb9a 100644
  # We compute the rest based on those here to avoid re-searching and to avoid finding a possibly
  # different installation.
 -if(CMAKE_CUDA_COMPILER_TOOLKIT_ROOT)
-+if(FALSE)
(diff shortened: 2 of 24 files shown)


jeongseok-meta:
Locally, I confirmed that manually applying this change to the installed `FindCUDAToolkit.cmake` (located in `<prefix>/share/cmake/Caffe2/`) resolves the following build error:

https://github.com/facebookincubator/momentum/actions/runs/12207717521/job/34059648061#step:8:3777

If this patch does not alter the pytorch build in this feedstock (while only addressing the downstream issue), I believe this can be a viable solution.

jeongseok-meta:

jeongseok-meta:
Also, I confirmed that #283 remains resolved from local build of pytorch (no "Unable to find cublas_v2.h in ..." messages). Therefore, I believe this patch addresses both issues: #283 and the downstream issue where cudatoolkit cannot be found.

recipe/build.sh
```diff
@@ -2,6 +2,9 @@
 
 set -ex
 
+# This is used to detect if it's in the process of building pytorch
+export IN_PYTORCH_BUILD
```
mgorny, line 6: Are you sure you didn't need to assign it some value?

mgorny (commented):
Thanks for doing this!

recipe/build.sh
jeongseok-meta: CMake, in a boolean context (e.g., if($ENV{MY_ENV})), an empty string is considered false. Let me assign a specific value to fix it. Thanks for catching this!

mgorny (approved):
WFM, thanks again!

h-vetinari:
Automerge on this feedstock is not such a good idea, because the bot has no rights to trigger the open-gpu server, and the commits by the bots also cannot be restarted manually. I'll push a restart.

jeongseok-meta:
Ah, it makes sense. Thank you for taking care of this!

recipe/patches/0010-Allow-overriding-CUDA-related-paths.patch
```diff
@@ -17,7 +17,7 @@ index ec9ae530..76d1fb9a 100644
  # We compute the rest based on those here to avoid re-searching and to avoid finding a possibly
  # different installation.
 -if(CMAKE_CUDA_COMPILER_TOOLKIT_ROOT)
-+if(FALSE)
++if(CMAKE_CUDA_COMPILER_TOOLKIT_ROOT AND NOT "$ENV{IN_PYTORCH_BUILD}" STREQUAL "1")
```
jakirkham, line 20: FWIW conda-build sets the environment variable `CONDA_BUILD=1`

ref: https://docs.conda.io/projects/conda-build/en/latest/user-guide/environment-variables.html#environment-variables-set-during-the-build-process

h-vetinari: Yeah, but that will be set in all the other downstream conda-forge builds as well, making it unusable to distinguish the situation between here (where we want to disable this part) and downstream (where we don't; at least not yet).

jakirkham: Good point

PR #299, opened 2024-12-04
DEBUG

h-vetinari:
~Demonstrate that fix in https://github.com/conda-forge/libcufile-feedstock/pull/24 was not sufficient, and that https://github.com/conda-forge/libcufile-feedstock/commit/733742d0a153591ca37ac2b597616937a3b965e5 needs to be reverted.~

Debugging run for #298 (which contains the commits for the original goal of this PR) + #304

PR #304, opened 2024-12-17
Fix dot in cross-compiling

isuruf:
Checklist


h-vetinari:
Thanks a lot! 🙏

I suggest we fold this into #298, which is basically complete and ready for merge (modulo this issue).

mgorny:
I'll test and merge it into #298.

h-vetinari:
The failures here
```
/Users/runner/miniforge3/conda-bld/libtorch_1734438514677/work/conda_build.sh: line 217: wheel: command not found
```
are likely fixed by https://github.com/conda-forge/pip-feedstock/pull/133

h-vetinari:
OK, this got picked into #298, but still fails for osx + generic there (though not for osx +MKL... 🤔; this is using an older MKL because intel stopped supporting osx-64). Additionally, linux + MKL failed as well (rest still running). None of these are cross-compiled, which is AFAIU why the upstream [logic](https://github.com/pytorch/pytorch/blob/v2.5.1/cmake/Modules/FindBLAS.cmake#L320-L384) for these variables never triggers, as it's guarded by `IF (CMAKE_CROSSCOMPILING)`.

hmaarrfk:
is it ok if we decouple it from the CUDA stuff in #298? Its pretty hard to keep both in mind, and they are very distinct ideas.

I think given the challenges of this repo, it is perfectly fine to attack the problems one at a time. In the debugging phase, which you seem to suggest we still are in, I think we can add skips to reduce the redundant load on the CIs.

I also think letting Isuru work on his own branch helps empower him to move faster instead of coming into conflicts with a larger effort.

recipe/meta.yaml
```diff
@@ -106,6 +106,7 @@ requirements:
     # other requirements
     - python 3.12  # [megabuild]
     - python       # [not megabuild]
+    - wheel
```
h-vetinari, line 109: not necessary; it was already fixed by the pip PR I mentioned

h-vetinari:
> is it ok if we decouple it from the CUDA stuff in #298? Its pretty hard to keep both in mind, and they are very distinct ideas.

Fine by me. I only advocated its inclusion because it seemed like a done deal... I've started a debugging job in #299 that adds two more commits on top to #298. If that doesn't yield anything, we can back out that commit and just skip the tests.

h-vetinari:
Since this PR does not include the test suite where this happens, we won't actually see whether this changes anything (which is why I based the run in #299 on #298). I'd suggest to cancel the run on the GPU server here, to not waste resources.

recipe/meta.yaml
isuruf: It's better to be explicit about the dependencies we use.

hmaarrfk:
this unfortunately needs to get rebased, and somebody needs to validate that this does work.

isuruf:
Merged in #298

PR #306, opened 2024-12-26
Prune torch cuda arch list to match upstream

isuruf:
Checklist

### DIFF
--- a/recipe/bld.bat
+++ b/recipe/bld.bat
@@ -16,7 +16,8 @@ if "%build_with_cuda%" == "" goto cuda_flags_end
 
 set CUDA_PATH=C:\Program Files\NVIDIA GPU Computing Toolkit\CUDA\v%desired_cuda%
 set CUDA_BIN_PATH=%CUDA_PATH%\bin
-set TORCH_CUDA_ARCH_LIST=5.0;6.0;6.1;7.0;7.5;8.0;8.6;8.9;9.0+PTX
+REM Keep this list in sync with https://github.com/pytorch/pytorch/blob/07fa6e2c8b003319f85a469307f1b1dd73f6026c/.ci/manywheel/build_cuda.sh#L60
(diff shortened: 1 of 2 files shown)


hmaarrfk:
It seems like a mistake to remove 8.9 to me. This targets 4090s and 6000 ada

isuruf:
Is there a difference between the code generated for 8.6 vs code generated for 8.9? Removing 8.9 here does not mean that those targets are dropped. Just that 8.9 compute capability devices will use code compiled for 8.6.

hmaarrfk:
> Is there a difference between the code generated for 8.6 vs code generated for 8.9? Removing 8.9 here does not mean that those targets are dropped. Just that 8.9 compute capability devices will use code compiled for 8.6.

I’m honestly really not sure and it is difficult to tell without further inspection. How can one tell in practice?

isuruf:
You have to run benchmarks to see if it does make sense. Since this is what upstream wheels do and most people use those wheels, I think we would have seen reports upstream if there was a difference.

hmaarrfk:
> I think we would have seen reports upstream if there was a difference.

I think very few people have skills to compile pytorch from source.

I can run a benchmark in 1 week if you can prepare one for me. I have a 4090.

hmaarrfk:
I think there are a few differences in the capabilities of 8.6 and 8.9 that we should keep both...
https://docs.nvidia.com/cuda/cuda-c-programming-guide/#arithmetic-instructions

they seem hard to trigger and would take us a lot of time to design benchmarks for (i think)

hmaarrfk:
It also seems that many lesser 40 series would benefit. Not just “those with highest budgets”

rgommers (approved):
Benchmarks are always great to inform decisions on what to build yes or no, so 👍🏼 from me for waiting for those. In principle this change seems correct though, and in the absence of evidence to the contrary of significant value of making different choices, it seems like the upstream `TORCH_CUDA_ARCH_LIST` should always be matched.

Issue #315, opened 2025-01-08
linux torch builds report an incompatible version string


bschindler:
### Solution to issue cannot be found in the documentation.

- [x] I checked the documentation.

### Issue

When installing pytorch from conda-forge, the version reported through importlib doesn't match the version installed on conda-forge, as in it contains a version suffix. In case of 2.4.1, the version reported is: 

2.4.1.post300

This causes a number of issues. In "simple" cases, it is that semver cannot parse this string: 
```
>>> import importlib.metadata
>>> v = importlib.metadata.version("torch")
>>> v
'2.4.1.post300'
>>> import semver
>>> semver.Version.parse(v, optional_minor_and_patch=True)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/opt/miniconda/envs/env/lib/python3.10/site-packages/semver/version.py", line 646, in parse
    raise ValueError(f"{version} is not valid SemVer string")
ValueError: 2.4.1.post300 is not valid SemVer string
```

It can also cause issues installing pip packages that depend on specific torch versions. For example, we ran into a case where the package requirements were torch >2.4,<=2.4.1. This fails with this version string. 

It would be fantastic if the conda package could report just 2.4.1 assuming this being feasible. 

Knowing that backports are hard, I am fine with this being fixed just for 2.5

### Installed packages

```shell
# packages in environment at /opt/miniconda/envs/torch:
#
# Name                    Version                   Build  Channel
_libgcc_mutex             0.1                 conda_forge    conda-forge
_openmp_mutex             4.5                  2_kmp_llvm    conda-forge
bzip2                     1.0.8                h4bc722e_7    conda-forge
c-ares                    1.34.4               hb9d3cd8_0    conda-forge
ca-certificates           2024.12.14           hbcca054_0    conda-forge
cpp-expected              1.1.0                hf52228f_0    conda-forge
cpython                   3.10.16         py310hd8ed1ab_1    conda-forge
cuda-cudart               12.6.77              h5888daf_0    conda-forge
cuda-cudart_linux-64      12.6.77              h3f2d84a_0    conda-forge
cuda-nvrtc                12.6.85              hbd13f7d_0    conda-forge
cuda-nvtx                 12.6.77              hbd13f7d_0    conda-forge
cuda-version              12.6                 h7480c83_3    conda-forge
cudnn                     9.3.0.75             h62a6f1c_2    conda-forge
[...]

PR #305, opened 2024-12-25
Reenable building kineto, add CUPTI dep

mgorny:
Checklist
* [x] Used a [personal fork of the feedstock to propose changes](https://conda-forge.org/docs/maintainer/updating_pkgs.html#forking-and-pull-requests)
* [ ] Bumped the build number (if the version is unchanged)
* [ ] Reset the build number to `0` (if the version changed)
* [x] [Re-rendered]( https://conda-forge.org/docs/maintainer/updating_pkgs.html#rerendering-feedstocks ) with the latest `conda-smithy` (Use the phrase <code>@<space/>conda-forge-admin, please rerender</code> in a comment in this PR for automated rerendering)
* [x] Ensured the license file is being packaged.

Fixes #76

Let's try enabling new dependencies separately, to see which one caused CI problems.

Issue #272, opened 2024-10-06
Use upstream's new split build


hmaarrfk:
### Comment:

It seems upstream introduced two new parameters to their build system:

* `BUILD_LIBTORCH_WHL`
* `BUILD_PYTHON_ONLY`

see: https://github.com/pytorch/pytorch/commit/a25b28a753775a57d22fd42be1d4c3975c5d4549

It seems that these made it to 2.4.0 and 2.4.1. It may be interesting to use these to alleviate some of the complexities of our existing build system.

However, these parameters seem very "pypi" centric where shared libraries are just shoved in the `site-packages` directory. We may want to massage things a little bit to make them work in a "conda" context.

cc: @baszalmstra

baszalmstra:
Appreciate you creating this issue. Ill dive into the issues with protobuf on Windows first but after that Id love to take a stab at this.

hmaarrfk:
yeah no problem. the fact that we have CIs that can run most of the linux + osx stuff might help somebody else jump in too!

rgommers:
Didn't @isuruf already implement a split build in this repo (xref gh-210)? I think the upstream split build came later. Is it a matter of simplifying @isuruf's original approach by relying on upstream support that's present now?

hmaarrfk:
> relying on upstream support that's present now?

yes.

mgorny:
I'd like to give this one a shot. For a start, I'm going to do it in place — but ideally, I think we would end up splitting the feedstocks, in order to be able to build them entirely separately. Is anyone opposed to that?

baszalmstra:
I wonder what the advantage of that is? Is it to decouple the two builds completely? I would assume most of the time they are quite coupled.

mgorny:
> I wonder what the advantage of that is? Is it to decouple the two builds completely? I would assume most of the time they are quite coupled.

It's primarily about build time (and space usage). Currently, building a single variant involves spending a lot of time building the common stuff ("libtorch"), and then a similar amount of time building PyTorch for all Python versions. From what I understand, splitting it would mean you can build libtorch first, and then use it to build PyTorch variants separately — effectively halving the time of each build.

baszalmstra:
But cant you achieve the same thing with the multi output recipe? Because from what I understand that is currently already the case

h-vetinari:
[...]

PR #316, opened 2025-01-10
Joint merge of windows and kineto PRs

h-vetinari:
To avoid running CI multiple times (as each run takes >1 day), merge #231 & #305, and pick up the pending fix for #315

Credit here goes largely to @baszalmstra @Tobias-Fischer @mgorny 👏


Tobias-Fischer:
Looks like we need “six” after all on osx @h-vetinari 

Thanks for the merge by the way!!

h-vetinari:
So aside from https://github.com/conda-forge/status/issues/189, we're running into a regression from an unexpected direction. OSX builds now time out 100%.

I suspect that the addition of `sccache` somehow causes `pytorch` to rebuild from scratch, rather than use the results of the global build step. In one of the few builds that got this far, I see
```
Resource usage statistics from bundling pytorch:
   Process count: 12
   CPU time: Sys=0:03:14.1, User=0:22:59.9
   Memory: 1.8G
   Disk usage: 5.7M
   Time elapsed: 2:39:53.5
```
Those extra ~2.5 hours are obviously a killer after
```
Resource usage statistics from building libtorch:
   Process count: 10
   CPU time: Sys=0:14:07.5, User=4:10:17.7
   Memory: 2.1G
   Disk usage: 5.1M
   Time elapsed: 2:46:00.6
```

Any ideas @Tobias-Fischer @mgorny?

mgorny:
Hmm, but sccache was only added on Windows?

recipe/meta.yaml
```diff
@@ -189,31 +247,34 @@ outputs:
...
         - llvm-openmp    # [osx]
+        - intel-openmp {{ mkl }}  # [win]
         - cmake
         - ninja
         # Keep libprotobuf here so that a compatibile version
         # of protobuf is installed between build and host
         - libprotobuf
         - protobuf
         - make      # [linux]
+        - sccache
```
h-vetinari, line 258: Here it was added unconditionally. Happy to add a `# [win]` selector of course.

mgorny, line 258: Oh, there's a missing `# [win]` here.

recipe/bld.bat
```diff
@@ -1,47 +1,221 @@
...
+  for /f "tokens=2" %%a in ('python --version 2^>^&1') do set PY_VERSION_FULL=%%a
 
-set DISTUTILS_USE_SDK=1
+  :: Replace Python312 or python312 with ie Python311 or python311
+  sed "s/\([Pp]ython\)312/\1%CONDA_PY%/g" build/CMakeCache.txt.orig > build/CMakeCache.txt
 
-set CMAKE_INCLUDE_PATH=%LIBRARY_PREFIX%\include
-set LIB=%LIBRARY_PREFIX%\lib;%LIB%
+  :: Replace version string v3.12.8() with ie v3.11.11()
+  sed -i.bak -E "s/v3\.12\.[0-9]+/v%PY_VERSION_FULL%/g" build/CMakeCache.txt
```
danpetry, line 47: fwiw we're using the builtin `BUILD_LIBTORCH_WHL` variables so we don't have to do all this. Seems to be working and testing well although I haven't checked in detail. https://github.com/AnacondaRecipes/pytorch-feedstock/blob/58b281e091816c0074e3754b31a8596f4e2ddaad/recipe/bld.bat#L111
If this is working, though, might be worth merging and looking at this later. I don't want to de-rail it when it's almost over the line. @ELundby45

mgorny: At least this didn't seem to work on Windows (hence sccache), but on Linux the difference is that doing the `sed` causes CMake to reuse most of the compiled files, but using re-cmake means it rebuilds a lot of common bits.

Tobias-Fischer: I’m not sure if we really ever tried their inbuilt approach - I haven’t but maybe @baszalmstra has?

baszalmstra: Nope, my attempts were all before BUILD_LIBTORCH_WHL was available. If it works without this “hack” it would be a big improvement because we can also get rid of the sccache workaround.

h-vetinari: Well, we have time to experiment a bit, currently the open-gpu-server is completely down and won't come back up before next week. So if someone wants to push into this PR (or you can try it out in #231 and I merge it here afterwards), feel free.

mgorny: The upstream options were discussed in #272. Long story short, I've tried and it was worse than status quo (the whole ATen tree was built multiple times).

hmaarrfk:
Just FYI, I'm about 4/6 builds through builds locally. I don't have a GPU but the tests are passing.

I've been building off b038c2d54d9b0cf6290acde96a7de95f5f3608a3

The fact that the windows build are passing and have gone through so many eyes should be reason enough to merge this. My builds should be completed in 12-18 hours I believe.

recipe/bld.bat
danpetry: yep ok, we're seeing similar things. thanks for the input. Ideally this mechanism could be fixed upstream, probably.

conda-forge.yml
```diff
@@ -20,4 +23,5 @@ github_actions:
 provider:
   linux_64: github_actions
```
hmaarrfk, line 24:
```suggestion
  linux_64: azure
```

I would like to merge this in, rerender, and merge immediately (to save the dedicated CI resources). We don't need to be so pedantic the CI building especially when many of us are looking at this. 

I can upload the builds to c-f when completed in 12-18 hours.

h-vetinari: > We don't need to be so pedantic the CI building especially when many of us are looking at this.

I agree, which is why I would just merge this as-is (thanks a lot for building out the linux side!).

The server outage is not a permanent thing, we were just unlucky with the Xmas break (GPU capacity was already reduced to half since December) and now the breakage on a Friday that took the rest offline. But this should be resolved sooner rather than later, and so a switch to azure (where the builds time out anyway) makes little sense to me.

hmaarrfk:
- 300MB uncompressed logs
- 18 MB [log_files.zip](https://github.com/user-attachments/files/18413288/log_files.zip)
- Could be 7MB xz in a zip [log_files_xz_ed.zip](https://github.com/user-attachments/files/18413336/log_files_xz_ed.zip)

should i upload directly to conda-forge?

conda-forge.yml
hmaarrfk: ok. as you wish.

hmaarrfk:
uploads should be completed within minutes

pgunn:
@hmaarrfk Did something go wrong? I see a bunch of new uploads, but none of them are windows.

h-vetinari:
Windows artefacts are still being built

pgunn:
@h-vetinari It's there now! Hooray!

h-vetinari:
Windows CUDA builds finished just now, should be through the CDN in ~20min

roaldarbol:
Just wanted to congratulate everyone involved in this PR! Terrific job everyone, it's not gone unnoticed! 🥳

h-vetinari:
> CDN is having issues right now

CUDA builds should also be downloadable now 🥳

jakirkham:
Congratulations!!! 🥳

This is a huge accomplishment. Thank you all for your hard work here 😄 👏

maresb:
I was so confused because my pytorch-dependent staged recipe was passing under Windows. Now I know why. :joy: 

Incredible work!!! :rocket: :partying_face:

PR #318, opened 2025-01-14
Sync some features with anaconda's recipe

danpetry:
Checklist

See commit messages for individual changes.
If there are any changes that aren't wanted, it'd be nice to know before debugging them.
There are some more involved changes we have over our side, and they will follow in further PRs if after investigation they're deemed appropriate. (full diff [here](https://github.com/AnacondaRecipes/pytorch-feedstock/pull/45), for interest)


recipe/build.sh
```diff
@@ -100,15 +120,20 @@ if [[ "${CI}" == "github_actions" ]]; then
     # cirun-openstack-gpu-2xlarge, which has 32GB RAM, 8 CPUs
     export MAX_JOBS=4
 else
-    export MAX_JOBS=${CPU_COUNT}
+    # Leave a spare core for other tasks. This may need to be reduced further
+    # if we get out of memory errors.
+    export MAX_JOBS=$((CPU_COUNT > 1 ? CPU_COUNT - 1 : 1))
```
jakirkham, line 125: What happens if `CPU_COUNT` is undefined or empty string?

recipe/build.sh
```diff
@@ -164,12 +189,19 @@ elif [[ ${cuda_compiler_version} != "None" ]]; then
...
+    # explicitly pass TORCH_CUDA_ARCH_LIST to cmake instead of implicitly
+    # setting it as an env variable.
+    #
+    # See:
+    # https://pytorch.org/docs/stable/cpp_extension.html (Compute capabilities)
+    # https://github.com/pytorch/pytorch/blob/main/.ci/manywheel/build_cuda.sh
     case ${cuda_compiler_version} in
         12.6)
             export TORCH_CUDA_ARCH_LIST="5.0;6.0;6.1;7.0;7.5;8.0;8.6;8.9;9.0+PTX"
             ;;
```
jakirkham, line 202: So what should we do differently to prepare for this?

recipe/meta.yaml
```diff
@@ -380,6 +399,29 @@ outputs:
...
+        - export MATRIX_GPU_ARCH_VERSION="none"                                                 # [(cuda_compiler_version == "None") and (not win)]
+        - export MATRIX_GPU_ARCH_TYPE="none"                                                    # [(cuda_compiler_version == "None") and (not win)]
+        - export MATRIX_CHANNEL="defaults"                                                      # [not win]
+        - export MATRIX_STABLE_VERSION="{{ version }}"                                          # [not win]
+        - export MATRIX_PACKAGE_TYPE="conda"                                                    # [not win]
+        - export TARGET_OS="linux"                                                              # [linux]
+        - export TARGET_OS="macos-arm64"                                                        # [(osx and arm64)]
+        - export TARGET_OS="macos-x86_64"                                                       # [(osx and x86_64)]
+        - export OMP_NUM_THREADS=4                                                              # [not win]
+        - python ./smoke_test/smoke_test.py --package torchonly
```
jakirkham, line 423: What happens if this runs without a GPU?

recipe/meta.yaml
```diff
@@ -380,6 +399,29 @@ outputs:
         # python-version-specific library (default location in SP_DIR symlinks back to this)
         - test -f $PREFIX/lib/libtorch_python${SHLIB_EXT}     # [unix]
 
+        # the smoke test script takes a bunch of env variables, defined below
```
jakirkham, line 402: Could we expand this comment a bit to describe the need for the env vars a bit more. There are quite a few

May also help to link out to some docs if there are good references available

jakirkham (commented):
Thanks Daniel! 🙏

Had a few questions below

danpetry:
The linter doesn't like this line:
```
- export MATRIX_GPU_ARCH_VERSION="{{ '.'.join(cuda_compiler_version.split('.')[:2]) }}" # [(cuda_compiler_version != "None") and (linux and x86_64)]
```
I guess it doesn't like calling .split on cuda_compiler_version. I'm getting `TypeError: 'NoneType' object is not callable` when rendering and `OSError: Feedstock has no recipe/meta.yaml` when linting. Deleting this line solves it, but I've tried various jinja stuff and can't solve it while keeping it in. Any ideas from someone who knows what the linter is looking for? If not I can continue...

jakirkham:
Normally something like this will do the trick

```
{{ ... (cuda_compiler_version or "").split(".") .. }}
```

recipe/build.sh
```diff
@@ -26,6 +26,10 @@ export USE_NUMA=0
 export USE_ITT=0
 
 #################### ADJUST COMPILER AND LINKER FLAGS #####################
+# Pytorch's build system doesn't like us setting the c++ standard and will
+# issue a warning:
+# https://github.com/pytorch/pytorch/blob/3beb7006dd5a415dfa236081ad5d55ae38346324/CMakeLists.txt#L41
+export CXXFLAGS="$(echo $CXXFLAGS | sed 's/-std=c++[0-9][0-9]//g')"
```
h-vetinari, line 32: We need to ensure that this is C++17 for reasons of abseil's ABI, see https://github.com/conda-forge/abseil-cpp-feedstock/issues/45. Luckily, pytorch 2.5 is on C++17 [already](https://github.com/pytorch/pytorch/blob/v2.5.1/CMakeLists.txt#L42) (and I agree with pytorch that this shouldn't be done through compiler flags but through `CMAKE_CXX_STANDARD`). Furthermore, conda-forge doesn't set the C++ standard in the compiler activation, so there's little chance that we see this in `CXXFLAGS`, but in any case, this change should be benign.

recipe/build.sh
h-vetinari: What CI provider do things run on for anaconda? I think it might need to become specific to that, and not in the overall `else:` clause here. Our OSX builds on azure work are successful withing the 6h limit, but seem to fail with this PR. This is a likely culprit.

recipe/patches/0016-use-prefix-include-for-inductor.patch
h-vetinari, line 1: Could you please generate this with `git format-patch ...`. We try to keep a consistent format, not least because then one can do something like `find ../path/to/patches/ | xargs git am`, which works with only git (no need to install something else locally), and has a windows-equivalent `for /f %f in ('dir /b /S ..\path\to\patches') do git am %f`

h-vetinari (commented):
Overall this looks very nice, thanks a lot. I left some questions on the individual patches.

recipe/meta.yaml
```diff
@@ -2,6 +2,9 @@
 {% set version = "2.5.1" %}
 {% set build = 9 %}
```
h-vetinari, line 3: We still need to bump this before merging.

recipe/build.sh
```diff
@@ -62,6 +79,7 @@ done
 CMAKE_FIND_ROOT_PATH+=";$SRC_DIR"
 unset CMAKE_INSTALL_PREFIX
 export TH_BINARY_BUILD=1
+# Use our build version and number for inserting into binaries
```
mgorny, line 82:
```suggestion
# Use our build version for inserting into binaries
```

We're *not* using the build number, as noted below.

recipe/build.sh
```diff
@@ -164,12 +191,19 @@ elif [[ ${cuda_compiler_version} != "None" ]]; then
             echo "unknown CUDA arch, edit build.sh"
             exit 1
     esac
+    # Warning from pytorch v1.12.1: In the future we will require one to
+    # explicitly pass TORCH_CUDA_ARCH_LIST to cmake instead of implicitly
+    # setting it as an env variable.
```
mgorny, line 196: Isn't that irrelevant to calling `setup.py`, though? After all, `setup.py` is passing it explicitly to CMake.

recipe/build.sh
```diff
@@ -204,15 +238,16 @@ case ${PKG_NAME} in
 
     mv build/lib.*/torch/bin/* ${PREFIX}/bin/
     mv build/lib.*/torch/lib/* ${PREFIX}/lib/
-    mv build/lib.*/torch/share/* ${PREFIX}/share/
+    # need to merge these now because we're using system pybind11, meaning the destination directory is not empty
+    rsync -a build/lib.*/torch/share/* ${PREFIX}/share/
```
mgorny, line 242: I would feel safer having an explicit `build` dependency on rsync. This doesn't look a tool one could reliably expect to be present on host.

mgorny:
@danpetry, BTW how do you feel about `rattler-build`? Because I'm eager to try converting PyTorch's recipe to that, and if we were to keep the two in sync, it'd probably be best to convert anaconda's as well.

jakirkham:
> @danpetry, BTW how do you feel about rattler-build? Because I'm eager to try converting PyTorch's recipe to that, and if we were to keep the two in sync, it'd probably be best to convert anaconda's as well.

This seems like a good item for discussion

Could I suggest we create an issue to discuss this topic explicitly?

danpetry:
Yep. Michal and I synced on this on zulip today. To summarise:
- as stated [here](https://hackmd.io/@conda-forge/HyypoJnLyl), we want to align but don't know timelines internally yet
- Michal seems ok with holding off
let's discuss further in the issue

danpetry:
https://github.com/conda-forge/pytorch-cpu-feedstock/issues/323

recipe/patches/0016-use-prefix-include-for-inductor.patch
```diff
@@ -0,0 +1,16 @@
...
+
+Index: pytorch/torch/_inductor/cpp_builder.py
+===================================================================
+--- pytorch.orig/torch/_inductor/cpp_builder.py	2024-12-16 15:16:47.074821258 -0600
++++ pytorch/torch/_inductor/cpp_builder.py	2024-12-16 15:17:33.922130106 -0600
+@@ -1055,6 +1055,7 @@
+         + python_include_dirs
+         + torch_include_dirs
+         + omp_include_dir_paths
++        + [os.getenv('CONDA_PREFIX') + '/include']
```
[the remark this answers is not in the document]
danpetry: inductor takes the compile_flags.json file for its AOT mode (handled in package.py), but not for its JIT mode. This is the problem the patch is solving - making it look in prefix/include during JIT (torch.compile) compilation. Probably in AOT mode the user wants to specify their own compile flags for their platform, which is what this json file is for.

The code you're looking at is where the AOT code is initializing the base class (BuildOptionsBase). However, we want to initialize the include directories in the child CppTorchOptions class, which is instantiated in cpu_vec_isa.py. It was a while ago I wrote this patch but IIRC that was the path in the stack trace.

recipe/build.sh
```diff
@@ -204,15 +236,16 @@ case ${PKG_NAME} in
...
     mv build/lib.*/torch/include/{ATen,caffe2,tensorpipe,torch,c10} ${PREFIX}/include/
     rm ${PREFIX}/lib/libtorch_python.*
 
     # Keep the original backed up to sed later
     cp build/CMakeCache.txt build/CMakeCache.txt.orig
     ;;
   pytorch)
-    $PREFIX/bin/python -m pip install . --no-deps -vvv --no-clean \
+    $PREFIX/bin/python -m pip install . --no-deps --no-build-isolation -vvv --no-clean \
```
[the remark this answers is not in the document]
danpetry: for windows we've got

`%PYTHON% -m pip %PIP_ACTION% . --no-build-isolation --no-deps -vvv --no-clean`

and unix

`$PREFIX/bin/python -m pip install . --no-deps --no-build-isolation -vvv --no-clean`

So the same AFAICS?
Keeping them here seems more in line with what people will expect given other feedstocks?

recipe/build.sh
```diff
@@ -74,6 +92,8 @@ export USE_SYSTEM_SLEEF=1
 # use our protobuf
 export BUILD_CUSTOM_PROTOBUF=OFF
 rm -rf $PREFIX/bin/protoc
+export USE_SYSTEM_PYBIND11=1
+export USE_SYSTEM_EIGEN_INSTALL=1
```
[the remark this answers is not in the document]
danpetry: It ignores them: [pybind11](https://github.com/pytorch/pytorch/blob/a8d6afb511a69687bbb2b7e88a3cf67917e1697e/cmake/Dependencies.cmake#L898) [eigen](https://github.com/pytorch/pytorch/blob/a8d6afb511a69687bbb2b7e88a3cf67917e1697e/cmake/Dependencies.cmake#L830)

Although in the case of eigen it will use the vendored one if it doesn't find the system one, rather than erroring, which can mean accidental vendoring if you don't look at the logs carefully, which isn't great IMHO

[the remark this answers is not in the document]
danpetry: we have a patch to prevent this for [openmp](https://github.com/AnacondaRecipes/pytorch-feedstock/blob/master/recipe/patches/0001-windows-FindMKL-add-library-suffix.patch) but I thought it best to look at the blas/openmp stuff closer and maybe address in a different PR.

recipe/build.sh
danpetry: It's AWS. Happy to just leave it as-is. I think originally we had it maxed at four cores and then increased it to CPU_COUNT-1 because of common practice and a desire to speed things up rather than concrete technical reasons. We can leave this as-is and if we have issues our end solve them later

recipe/build.sh
danpetry: probably worth attending to this now, you're right, will have a look

recipe/meta.yaml
danpetry: the env variables above tell the smoke test whether to test using a gpu or a cpu, so without a gpu it's fine

recipe/build.sh
danpetry: ok, fair. Anaconda does set the standard via `CXXFLAGS` but also sets it to 17. I think I had some issues with this in the past so it's probably safest to keep it in, but if you prefer us to keep it on our side I'm fine with that.

recipe/patches/0016-use-prefix-include-for-inductor.patch
danpetry: yes will do

recipe/build.sh
danpetry: I seem to remember seeing the error, also via setup.py, will investigate further

danpetry:
Have addressed some comments, will continue later

recipe/build.sh
h-vetinari: I've fixed our case in 6074128b18c469590ff29fb9485f4539daf86386, so I have no objection to doing what helps you in the `else:` branch.

danpetry: have made this a default in the else branch

jakirkham: Note that [`CPU_COUNT` is a passthrough environment variable]( https://github.com/conda/conda-build/blob/8df6493565dae517165754bd75d7d74de1522dc3/conda_build/environ.py#L794 )

So one can set this in their CI scripts to the intended value and conda-build will use it. This is [what conda-forge does]( https://github.com/conda-forge/conda-forge-ci-setup-feedstock/blob/d8b11fd85622f4d332cd099d91e8fa12562c9be6/recipe/run_conda_forge_build_setup_linux#L20-L28 )

Instead of adding this to the build recipe, would recommend doing this in Anaconda's CI scripts

recipe/patches/0016-use-prefix-include-for-inductor.patch
danpetry: done

recipe/build.sh
danpetry: done

recipe/build.sh
danpetry: I don't think this variable gets passed via setup.py, it gets detected in a cmake module https://github.com/pytorch/pytorch/blob/266fd35c5842902f6304aa8e7713b252cbfb243c/cmake/public/utils.cmake#L327

Setting it via CMAKE_ARGS should solve this, have changed it to that

recipe/build.sh
danpetry: done

recipe/build.sh
danpetry: left as-is

recipe/meta.yaml
danpetry: done

recipe/build.sh
danpetry: have handed the variable via CMAKE_ARGS

recipe/build.sh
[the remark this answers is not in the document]
danpetry: are you ok with this?

recipe/build.sh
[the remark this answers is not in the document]
danpetry: any further comments?

recipe/patches/0016-use-prefix-include-for-inductor.patch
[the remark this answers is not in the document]
danpetry: does this check out ok for you?

danpetry:
Changes addressed, will debug further

recipe/meta.yaml
```diff
@@ -392,6 +437,7 @@ outputs:
             "test/test_nn.py",
             "test/test_torch.py",
             "test/test_xnnpack_integration.py",
+            "test/inductor/test_torchinductor.py",                                      # [cuda_compiler_version != "None"]
             "-m \"not hypothesis\"",
         ]) %}
```
h-vetinari, line 442: Unless something changed recently, it's not possible to use selectors in jinja expressions. But it's possible to do
```suggestion
            "test/test_xnnpack_integration.py",
        ] + (cuda_compiler_version != "None") * ["test/inductor/test_torchinductor.py"]) %}
```
(and then either move `-m \"not hypothesis\"` to the front of the list - or even better, to the pytest invocation).

recipe/build.sh
danpetry: `CPU_COUNT` isn't being set here. If it's set (somewhere else, like in CI as you suggest) then it'll be used to set `MAX_JOBS`. If it's not set, this expression will evaluate to 1

recipe/meta.yaml
```diff
@@ -376,7 +376,7 @@ outputs:
         # tools/ is needed to optimise test run
         # as of pytorch=2.0.0, there is a bug when trying to run tests without the tools
         - tools
-        - .ci/pytorch/smoke_test/smoke_test.py
+        #- .ci/pytorch/smoke_test/smoke_test.py
```
h-vetinari, line 379: Looks like https://github.com/pytorch/pytorch/commit/f57606ab850bd304cfb4a16f8ac62cd66fe23bbe landed only for 2.6.0, which is around the corner, so this still makes sense to keep at least.

danpetry: I was going to ask if you wanted this on the 2.6.0 PR; here, it blocks the rest of the testing because this is still on 2.5.1

h-vetinari: logically it makes more sense to add this in the 2.6 PR, but I don't really mind if it lands (though deactivated) in this PR already

danpetry:
Does conda-forge's CI cross-compile osx-arm64 packages on an osx-64 platform? I'm getting the following error (on osx-arm64) and can't recreate locally (on an arm mac, either with or without x86_64 emulation). Alternatively, if anyone's seen this error and knows what's going on, that would be helpful. I think it might be something to do with the cmake_args decomposing/exporting block that's been deleted.

```
[10/6853] Building C object confu-deps/cpuinfo/CMakeFiles/cpuinfo.dir/src/x86/init.c.o
FAILED: confu-deps/cpuinfo/CMakeFiles/cpuinfo.dir/src/x86/init.c.o 
$BUILD_PREFIX/bin/arm64-apple-darwin20.0.0-clang -DCPUINFO_LOG_LEVEL=2 -DCPUINFO_LOG_TO_STDIO=1 -I$SRC_DIR/third_party/cpuinfo/src -I$SRC_DIR/third_party/cpuinfo/include -ftree-vectorize -fPIC -fstack-protector-strong -O2 -pipe -isystem $PREFIX/include -fdebug-prefix-map=$SRC_DIR=/usr/local/src/conda/libtorch-2.5.1 -fdebug-prefix-map=$PREFIX=/usr/local/src/conda-prefix -Wno-deprecated-declarations -Wno-unknown-warning-option -Wno-error=unused-command-line-argument -Wno-error=vla-cxx-extension -O3 -DNDEBUG -std=c99 -isysroot /Applications/Xcode_15.2.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX13.3.sdk -mmacosx-version-min=11.0 -fPIC -MD -MT confu-deps/cpuinfo/CMakeFiles/cpuinfo.dir/src/x86/init.c.o -MF confu-deps/cpuinfo/CMakeFiles/cpuinfo.dir/src/x86/init.c.o.d -o confu-deps/cpuinfo/CMakeFiles/cpuinfo.dir/src/x86/init.c.o -c $SRC_DIR/third_party/cpuinfo/src/x86/init.c
In file included from $SRC_DIR/third_party/cpuinfo/src/x86/init.c:9:
In file included from $SRC_DIR/third_party/cpuinfo/src/x86/cpuid.h:5:
$BUILD_PREFIX/lib/clang/18/include/cpuid.h:14:2: error: this header is for x86 only
```

h-vetinari:
> Does conda-forge's CI cross-compile osx-arm64 packages on an osx-64 platform?

Yes, we don't have native arm runners at scale, so we're forced to cross-compile



> I think it might be something to do with the cmake_args decomposing/exporting block that's been deleted.

💯

danpetry:
I've reverted some commits that are taking more time than I'd like and will address them in further PRs if that's ok.

danpetry:
does the CI stop previous jobs automatically if you push a new commit?
Relatedly, I was wondering what's the best place to find info on conda-forge's CI, the rendering process, and other development flow/infrastructure stuff? Is it [here](https://conda-forge.org/docs/maintainer/infrastructure/)?

recipe/meta.yaml
danpetry: As an aside, for further development flow optimization, we could run just the smoke tests on non-release PRs.

h-vetinari: What kind of runtime are we expecting the smoketest to take? If it's <30min, it's fine to run always IMO.

danpetry:
We can expect a test failure on Windows, by the way, detailed in [this](https://github.com/AnacondaRecipes/pytorch-feedstock/issues/55#issue-2810442549) issue. It's due to a test 
(which defines a custom operator to do some JIT compilation with pybind11) not finding the pybind11 headers. The reason they aren't found is most likely because by default the pytorch code looks for the headers in the pip paths but not the conda paths (preceded by `Library`).

The fix we had was 
`mv %PREFIX%/include/pybind11_global/pybind11 %PREFIX%/include  # [win]`
but decided to remove it, document the failure and skip the test, as a) it's pretty edge casey and b) the writer of the custom operator can specify `extra_include_paths`. =

I'm just letting you know before the Windows builds run. Do you think it should be fixed or are you ok with skipping it?

recipe/meta.yaml
danpetry: it's pretty brief, much briefer than the suite of tests. but what I mean is: for non-release PRs, only run the smoke tests, don't run the rest of the tests. Then for releases, run all the tests.

h-vetinari:
> I'm just letting you know before the Windows builds run. Do you think it should be fixed or are you ok with skipping it?

I expect to merge #328 first, where we already point the include paths of parts of the test infrastructure towards `%LIBRARY_INC%` - ideally this solves the pybind issue already, and if not it's probably worth fixing (it falls into the same category as https://github.com/conda-forge/pytorch-cpu-feedstock/pull/328/commits/e0e34e37edb8ae492e15129693e5e1c01c282070, which also fixed "only" 2 tests).

After merging #328, I'll merge back main here, and then we can get this over the finish line!

recipe/meta.yaml
h-vetinari: What's a "non-release PR"? One for a release candidate? Or what do you mean there?

danpetry: I mean a PR where features are added but upon merge, a new conda package won't be uploaded to conda-forge. "release" from the point of view of the conda package release.

h-vetinari: OK, that's what confused me, because we never release anything based on what's built in a PR, it needs to be merged anyway. And I don't agree to not run the full tests in PRs, that just means we get negative surprises _after_ merging (and the builds that passed ended up being published anyway, creating an inconsistent situation across build combination).

danpetry: > and the builds that passed ended up being published anyway, creating an inconsistent situation across build combination

I think there's something I don't understand about the workflow here too. are the builds for all merged PRs published? I think the answer is yes.

h-vetinari: Yes, every merge to `main` will publish builds. This is a core constraint in our setup, we don't have a switch that says "merge to main but don't publish" (well except from manually adding a `skip: true` in the recipe).

h-vetinari: Correction: such a switch does exist:
https://github.com/conda-forge/pytorch-cpu-feedstock/blob/4ff92b330141d5d496afdc2d83ca0f32f6adf03d/.scripts/build_steps.sh#L91-L93
but I've never seen `UPLOAD_PACKAGES` used.

h-vetinari:
I've merged main after #331 got in, but cancelled CI for now while we're building packages on `main`. Feel free to layer more changes on top of course!

I've applied the moral equivalent of https://github.com/conda-forge/pytorch-cpu-feedstock/commit/7f0e31db643d17739546a147227eff56c4f8f5be to your changes in `bld.bat` as well. Personally I also prefer `::`, but that can't be used in indented blocks without warnings, so I didn't feel like arguing about this. 😅 

If you disagree there, I'll let you hash it out with @mgorny ;-)

h-vetinari:
After the windows CPU build just finished on main, started one run of that job here

h-vetinari:
Not sure why the linux jobs here aren't starting... The server is up and running and has resources available. CC @aktech

Tobias-Fischer:
There’s a typo: https://github.com/conda-forge/pytorch-cpu-feedstock/blob/fb0f118fa15c822a15e2a3931c777444357fb559/recipe/conda_build_config.yaml#L27

h-vetinari:
> There’s a typo:

Gah, thanks.

h-vetinari:
This seems to have some new failures for the CUDA builds on linux:
```
=========================== short test summary info ============================
FAILED [0.0691s] test/test_nn.py::TestNN::test_grid_sample - RuntimeError: CUDA error: out of memory
CUDA kernel errors might be asynchronously reported at some other API call, so the stacktrace below might be incorrect.
For debugging consider passing CUDA_LAUNCH_BLOCKING=1
Compile with `TORCH_USE_CUDA_DSA` to enable device-side assertions.


    python test/test_nn.py TestNN.test_grid_sample

FAILED [0.0084s] test/inductor/test_torchinductor.py::CpuTests::test_mutable_custom_op_fixed_layout2_cpu - UnboundLocalError: cannot access local variable 'lib' where it is not associated with a value

    python test/inductor/test_torchinductor.py CpuTests.test_mutable_custom_op_fixed_layout2_cpu

FAILED [0.0044s] test/inductor/test_torchinductor.py::CpuTests::test_mutable_custom_op_fixed_layout_cpu - UnboundLocalError: cannot access local variable 'lib' where it is not associated with a value

    python test/inductor/test_torchinductor.py CpuTests.test_mutable_custom_op_fixed_layout_cpu

FAILED [0.0042s] test/inductor/test_torchinductor.py::GPUTests::test_mutable_custom_op_fixed_layout2_cuda - UnboundLocalError: cannot access local variable 'lib' where it is not associated with a value

    python test/inductor/test_torchinductor.py GPUTests.test_mutable_custom_op_fixed_layout2_cuda

FAILED [0.0044s] test/inductor/test_torchinductor.py::GPUTests::test_mutable_custom_op_fixed_layout_cuda - UnboundLocalError: cannot access local variable 'lib' where it is not associated with a value

    python test/inductor/test_torchinductor.py GPUTests.test_mutable_custom_op_fixed_layout_cuda

FAILED [0.2764s] test/inductor/test_torchinductor.py::TritonCodeGenTests::test_sdpa_inference_mode_aot_compile - torch._dynamo.exc.SkipFrame: BF16 is not supported

    python test/inductor/test_torchinductor.py TritonCodeGenTests.test_sdpa_inference_mode_aot_compile

= 6 failed, 14539 passed, 2666 skipped, 91 xfailed, 143305 warnings in 4381.93s (1:13:01) =
```
In more detail:
```
_______________ CpuTests.test_mutable_custom_op_fixed_layout_cpu _______________
[gw0] linux -- Python 3.11.11 $PREFIX/bin/python

args = ('mylib', 'DEF'), kwargs = {}

    @contextlib.contextmanager
    def _scoped_library(*args, **kwargs):
        try:
>           lib = Library(*args, **kwargs)

$PREFIX/lib/python3.11/site-packages/torch/library.py:408: 

self = <[AttributeError("'Library' object has no attribute 'kind'") raised in repr()] Library object at 0x7f779c990d50>
ns = 'mylib', kind = 'DEF', dispatch_key = ''

    def __init__(self, ns, kind, dispatch_key=""):
        if kind not in ("IMPL", "DEF", "FRAGMENT"):
            raise ValueError("Unsupported kind: ", kind)
    
        if ns in _reserved_namespaces and (kind == "DEF" or kind == "FRAGMENT"):
            raise ValueError(
                ns,
                " is a reserved namespace. Please try creating a library with another name.",
            )
    
        frame = traceback.extract_stack(limit=3)[0]
        filename, lineno = frame.filename, frame.lineno
>       self.m: Optional[Any] = torch._C._dispatch_library(
            kind, ns, dispatch_key, filename, lineno
        )
E       RuntimeError: Only a single TORCH_LIBRARY can be used to register the namespace mylib; please put all of your definitions in a single TORCH_LIBRARY block.  If you were trying to specify implementations, consider using TORCH_LIBRARY_IMPL (which can be duplicated).  If you really intended to define operators for a single namespace in a distributed way, you can use TORCH_LIBRARY_FRAGMENT to explicitly indicate this.  Previous registration of TORCH_LIBRARY was registered at /home/conda/.cache/torch_extensions/py311_cu126/mylib/main.cpp:23; latest registration was registered at /dev/null:137

$PREFIX/lib/python3.11/site-packages/torch/library.py:87: RuntimeError

During handling of the above exception, another exception occurred:

self = <inductor.test_torchinductor.CpuTests testMethod=test_mutable_custom_op_fixed_layout_cpu>

    @config.patch(implicit_fallbacks=True)
    def test_mutable_custom_op_fixed_layout(self):
>       with torch.library._scoped_library("mylib", "DEF") as lib:

test/inductor/test_torchinductor.py:10684: 
$PREFIX/lib/python3.11/contextlib.py:137: in __enter__
    return next(self.gen)

args = ('mylib', 'DEF'), kwargs = {}

    @contextlib.contextmanager
    def _scoped_library(*args, **kwargs):
        try:
            lib = Library(*args, **kwargs)
            yield lib
        finally:
>           lib._destroy()
E           UnboundLocalError: cannot access local variable 'lib' where it is not associated with a value
E           
E           To execute this test, run the following from the base repo dir:
E               python test/inductor/test_torchinductor.py CpuTests.test_mutable_custom_op_fixed_layout_cpu
E           
E           This message can be suppressed by setting PYTORCH_PRINT_REPRO_ON_FAILURE=0

$PREFIX/lib/python3.11/site-packages/torch/library.py:411: UnboundLocalError
```

h-vetinari:
Also no idea why the CPU builds didn't start. Perhaps I messed something up (in addition to the typo)? In any case, feel free to force-reset back to https://github.com/conda-forge/pytorch-cpu-feedstock/pull/318/commits/e19e11c0a57299a1b3f5f380bef6e3ffba2bc5cc (though that won't affect the failures above).

recipe/build.sh
```diff
@@ -4,6 +4,10 @@ echo "=== Building ${PKG_NAME} (py: ${PY_VER}) ==="
 
 set -ex
 
+echo "####################################################################"
+echo "Building PyTorch using BLAS implementation: $blas_impl              "
+echo "####################################################################"
```
mgorny, line 9: Nit: could we combine this with the message above somehow, so we get a single big message with all important details?

recipe/build.sh
```diff
@@ -98,16 +117,23 @@ if [[ "${CI}" == "github_actions" ]]; then
...
     # Fake openblas
     export BLAS=OpenBLAS
     export OpenBLAS_HOME=${PREFIX}
-else
+elif [[ "$blas_impl" == "mkl" ]]; then
     export BLAS=MKL
+else
+    echo "[ERROR] Unsupported BLAS implementation '${blas_impl}'" >&2
+    exit 1
 fi
```
mgorny, line 137: This looks like a case for a `case` now.

recipe/build.sh
danpetry: done

recipe/build.sh
danpetry: done

danpetry:
hmm, there's a bunch of infra stuff/discussion on this going on I see... rather than catching up on that I'll have a look at those CUDA-related failures

recipe/conda_build_config.yaml
```diff
@@ -24,7 +24,6 @@ blas_impl:
 github_actions_labels:          # [linux or win]
 - cirun-openstack-cpu-2xlarge   # [linux]
 - cirun-openstack-gpu-2xlarge   # [linux]
-- cirun-openstack-gpu-2xlarge   # [linux]
```
h-vetinari: you need to delete both 2/3 if you want to use the current pinning (post revert of https://github.com/conda-forge/conda-forge-pinning-feedstock/pull/6910)

danpetry: just run all linux builds on a gpu machine?

danpetry:
win and linux-aarch64 are failing on `torch.cuda.is_available()`, I'm assuming because they're running on a cpu-only machine?

recipe/conda_build_config.yaml
h-vetinari: Yeah, that's what I wanted to avoid, but we're stuck with it for now.

h-vetinari:
> win and linux-aarch64 are failing on `torch.cuda.is_available()`, I'm assuming because they're running on a cpu-only machine?

Cannot speak for the windows server, but I'd assume it doesn't have a GPU. @wolfv?

For linux-aarch64 there should be a GPU there, but not sure if that's compatible with emulating things through QEMU.

Since the `torch.cuda.is_available()` tests are new in this PR, I'd just skip them on win & aarch for now.

danpetry:
hmm, nope, at least for aarch64 there's
```
025-01-27T05:23:37.7647163Z github_actions_labels:
2025-01-27T05:23:37.7647730Z - cirun-openstack-gpu-2xlarge
```

danpetry:
> the torch.cuda.is_available() tests are new in this PR

so they are

danpetry:
`test/inductor/test_torchinductor.py::TritonCodeGenTests::test_sdpa_inference_mode_aot_compile - torch._dynamo.exc.SkipFrame: BF16 is not supported`
This test needs compute capability 8 and above. There's also an OOM error in `test_grid_sample` which could be because meta's CI runs on higher-resource GPUs than conda-forge's.
Assuming upgrading GPUs isn't an option, we could skip?

recipe/build.sh
danpetry: I was seeing `unused variables: CMAKE_ARGS` in the config for some platforms, so I'm not confident this is working. Will add a note to address this at a later date unless you object. I tried a couple of ways to hand this directly to CMake but given we're using setup.py as an entry point it's nontrivial

danpetry:
The logs are looking truncated. @h-vetinari are the logs showing the correct behaviour given your changes? Not sure whether to restart, or wait, given that you're making changes here too...

h-vetinari:
> The logs are looking truncated. @h-vetinari are the logs showing the correct behaviour given your changes?

The linux logs being truncated is expected, there was some reshuffling of the physical GPUs in the context of https://github.com/conda-forge/status/issues/189, however the good news is that this should be fixed now, and I expect a rerun to encounter no infrastructure issues anymore.

The failure on win+CUDA is genuine though, and looks like
```
=========================== short test summary info ===========================
FAILED [0.0274s] test/inductor/test_torchinductor.py::CpuTests::test_mutable_custom_op_fixed_layout_cpu - UnboundLocalError: cannot access local variable 'lib' where it is not associated with a value

    python test\inductor\test_torchinductor.py CpuTests.test_mutable_custom_op_fixed_layout_cpu

= 1 failed, 8119 passed, 1498 skipped, 43 deselected, 31 xfailed, 76576 warnings in 3480.33s (0:58:00) =
```

This looks like exactly the same error like we got on the linux+CUDA builds [before](https://github.com/conda-forge/pytorch-cpu-feedstock/pull/318#issuecomment-2615646545), so I expect this is still something we have to fix. It's based on the new inductor tests you've added, so I'm open to skipping those tests.

Looking more closely, the [test](https://github.com/pytorch/pytorch/blob/v2.5.1/test/inductor/test_torchinductor.py#L10683) in question fails to create some intermediate library:
```
================================== FAILURES ===================================
______________ CpuTests.test_mutable_custom_op_fixed_layout_cpu _______________

args = ('mylib', 'DEF'), kwargs = {}

    @contextlib.contextmanager
    def _scoped_library(*args, **kwargs):
        try:
>           lib = Library(*args, **kwargs)

..\_test_env\Lib\site-packages\torch\library.py:408: 

self = <[AttributeError("'Library' object has no attribute 'kind'") raised in repr()] Library object at 0x22d66c30110>
ns = 'mylib', kind = 'DEF', dispatch_key = ''

    def __init__(self, ns, kind, dispatch_key=""):
        if kind not in ("IMPL", "DEF", "FRAGMENT"):
            raise ValueError("Unsupported kind: ", kind)
    
        if ns in _reserved_namespaces and (kind == "DEF" or kind == "FRAGMENT"):
            raise ValueError(
                ns,
                " is a reserved namespace. Please try creating a library with another name.",
            )
    
        frame = traceback.extract_stack(limit=3)[0]
        filename, lineno = frame.filename, frame.lineno
>       self.m: Optional[Any] = torch._C._dispatch_library(
            kind, ns, dispatch_key, filename, lineno
        )
E       RuntimeError: Only a single TORCH_LIBRARY can be used to register the namespace mylib; please put all of your definitions in a single TORCH_LIBRARY block.  If you were trying to specify implementations, consider using TORCH_LIBRARY_IMPL (which can be duplicated).  If you really intended to define operators for a single namespace in a distributed way, you can use TORCH_LIBRARY_FRAGMENT to explicitly indicate this.  Previous registration of TORCH_LIBRARY was registered at C:\Users\runnerx.cirun-8e2044a\AppData\Local\torch_extensions\torch_extensions\Cache\py312_cu126\mylib\main.cpp:23; latest registration was registered at /dev/null:137

..\_test_env\Lib\site-packages\torch\library.py:87: RuntimeError

During handling of the above exception, another exception occurred:

self = <inductor.test_torchinductor.CpuTests testMethod=test_mutable_custom_op_fixed_layout_cpu>

    @config.patch(implicit_fallbacks=True)
    def test_mutable_custom_op_fixed_layout(self):
>       with torch.library._scoped_library("mylib", "DEF") as lib:
```
This bit
```
 Previous registration of TORCH_LIBRARY was registered at [...]\torch_extensions\Cache\py312_cu126\mylib\main.cpp:23; latest registration was registered at /dev/null:137
```
looks like there's some failure to clean resources from previous tests?

h-vetinari:
Looking at the state of that test in the upcoming 2.6, I noticed https://github.com/pytorch/pytorch/commit/381213ee8a62c3fe4bec4ffd22a83c19eb88dc96 & https://github.com/pytorch/pytorch/commit/8fa0479dd80359350f954602c4fabfb056dabbe1. Let me try backporting those

recipe/meta.yaml
```diff
@@ -476,6 +476,9 @@ outputs:
         {% set skips = skips ~ " or test_BCELoss_weights_no_reduce_cuda" %}             # [unix and cuda_compiler_version != "None"]
         {% set skips = skips ~ " or test_ctc_loss_cudnn_tensor_cuda " %}                # [unix and cuda_compiler_version != "None"]
         {% set skips = skips ~ " or (TestTorch and test_index_add_correctness)" %}      # [unix and cuda_compiler_version != "None"]
+        # These tests require higher-resource or more recent GPUs than the CI provides
+        {% set skips = skips ~ " or (TritonCodeGenTests and test_sdpa_inference_mode_aot_compile)" %}   # [unix and cuda_compiler_version != "None"]
+        {% set skips = skips ~ " or (TestNN and test_grid_sample)" %}                                   # [unix and cuda_compiler_version != "None"]
```
h-vetinari, line 481: OK, I was just redownloading the logs to double-check why they failed. But was going to do the same.

danpetry: we can also just skip those other tests and add a note if the backport doesn't work

h-vetinari: The backport was easy, but I was double-checking the remaining failures before pushing.

danpetry: or patch to rename the lib each time
```
def test_mutable_custom_op_fixed_layout(self):
    unique_lib_name = f"mylib_{id(self)}"  # Make unique name using test instance id
    with torch.library._scoped_library(unique_lib_name, "DEF") as lib:
```
Although I'm surprised I can't find anything about these tests failing in issues upstream.

h-vetinari:
I think this PR is getting very close to merging. We're down to a tiny handful of test failures, which will hopefully be addressed by the lastest round of pushes. Beyond that, I don't want to shove more feature work into this PR, tensorpipe support on windows (etc.) can come in another PR - the CMake fixes were unavoidable though...

Any other thoughts @danpetry?

danpetry:
I'm good with that

h-vetinari:
PS. So glad to see the GPU server back in full force - 4 simultaneous GPU jobs 🤩 - big thank you to @aktech for sorting out some very gnarly hardware issues there! 🙏 🚀

danpetry:
May I ask how you found that PR? By looking at these tests in upstream's main and seeing if anything's changed, suggesting a fix..?

h-vetinari:
> May I ask how you found that PR? By looking at these tests in upstream's main and seeing if anything's changed, suggesting a fix..?

Going through the blame of the test file. It's a bit cumbersome because you need to do it in a local checkout - the file's too big for GH to load the blame. Example (on `v2.6.0-rc9`, the line numbers on main have shifted again already)
```
>git blame -L 11030,11050 -- test/inductor/test_torchinductor.py
afabed6ae608 (James Wu       2024-01-21 07:06:31 -0800 11030)         with torch.no_grad():
afabed6ae608 (James Wu       2024-01-21 07:06:31 -0800 11031)             # With keep_output_stride False, inductor would normally have different layout from eager execution
afabed6ae608 (James Wu       2024-01-21 07:06:31 -0800 11032)             # But because our custom op needs fixed layout, the assertions in the custom op will pass
afabed6ae608 (James Wu       2024-01-21 07:06:31 -0800 11033)             self.common(fn, (inp,), check_lowp=False)
afabed6ae608 (James Wu       2024-01-21 07:06:31 -0800 11034)
afd081c9d4d4 (rzou           2024-08-21 10:08:01 -0700 11035)     @requires_gpu()
afd081c9d4d4 (rzou           2024-08-21 10:08:01 -0700 11036)     @config.patch(implicit_fallbacks=True)
381213ee8a62 (Benjamin Glass 2024-11-26 16:05:19 +0000 11037)     @skip_if_cpp_wrapper(
381213ee8a62 (Benjamin Glass 2024-11-26 16:05:19 +0000 11038)         "Without major redesign, cpp_wrapper will not support custom ops that are "
381213ee8a62 (Benjamin Glass 2024-11-26 16:05:19 +0000 11039)         "defined in Python."
381213ee8a62 (Benjamin Glass 2024-11-26 16:05:19 +0000 11040)     )
05cb98f91d49 (Eddie Yan      2024-10-30 20:34:14 +0000 11041)     @tf32_on_and_off(0.005)
afd081c9d4d4 (rzou           2024-08-21 10:08:01 -0700 11042)     def test_mutable_custom_op_fixed_layout2(self):
afd081c9d4d4 (rzou           2024-08-21 10:08:01 -0700 11043)         with torch.library._scoped_library("mylib", "DEF") as lib:
afd081c9d4d4 (rzou           2024-08-21 10:08:01 -0700 11044)             mod = nn.Conv2d(3, 128, 1, stride=1, bias=False).to(device=GPU_TYPE)
afd081c9d4d4 (rzou           2024-08-21 10:08:01 -0700 11045)             inp = torch.rand(2, 3, 128, 128, device=GPU_TYPE)
afd081c9d4d4 (rzou           2024-08-21 10:08:01 -0700 11046)             expected_stride = mod(inp).clone().stride()
afd081c9d4d4 (rzou           2024-08-21 10:08:01 -0700 11047)
afd081c9d4d4 (rzou           2024-08-21 10:08:01 -0700 11048)             lib.define(
afd081c9d4d4 (rzou           2024-08-21 10:08:01 -0700 11049)                 "bar(Tensor x, bool is_compiling) -> Tensor",
f65a564fa247 (rzou           2024-09-09 07:30:54 -0700 11050)                 tags=torch.Tag.flexible_layout,
```
From there you find the commit https://github.com/pytorch/pytorch/commit/381213ee8a62c3fe4bec4ffd22a83c19eb88dc96 (and its PR). Since that wasn't the one that introduced `skip_if_cpp_wrapper` though - go to the parent of that commit, find the line number of the test in the file again, and do another `git blame` around those lines to find https://github.com/pytorch/pytorch/commit/8fa0479dd80359350f954602c4fabfb056dabbe1 and its PR.

danpetry:
ok, thanks. Additionally, I see that cpp_wrapper should be [off by default](https://github.com/pytorch/pytorch/blob/a8d6afb511a69687bbb2b7e88a3cf67917e1697e/torch/_inductor/config.py#L76), which doesn't match up with the suggestion of the changes that these tests are failing because cpp_wrapper's enabled. Or, are you just taking these changes because they solve the issue by skipping and the patch can be removed with v2.6?

h-vetinari:
> Or, are you just taking these changes because they solve the issue by skipping and the patch can be removed with v2.6?

TBH I was looking for changes in the failing tests that would be plausible for fixing things, and we get to drop backported patches in the next version upgrade.

I didn't research much further (those rabbit holes can be an enormous time sink, so I went with 🤞 for now), and it's entirely possible that the fix is not sufficient. In that case I'm happy to skip the tests, but as that added skip was staring me in the face, I thought it's worth a shot. 🤷

danpetry:
> those rabbit holes can be an enormous time sink

yep. Thanks for the knowledge/process share.

h-vetinari:
Hm. We're failing the new CMake tests for the CUDA-enabled pytorch. Let me add a `{{ compiler("cuda") }}` to the test section.

h-vetinari:
Yay, more failures
```
+ python -c 'import torch; assert torch.backends.cudnn.is_available()'
+ python -c 'import torch; assert torch.cuda.is_available()'
  File "<string>", line 1, in <module>
AssertionError
```
Presumably also `pytorch` needs to find something CUDA-related to pass this check?

The test skip was also not effective, as seen on windows. I'll switch to your suggestion with the unique names.

Finally, we still have a problem with the CMake metadata, because the Caffe2 files still use the long-deprecated `find_package(CUDA)`. I'll comment in #333

danpetry:
> Presumably also pytorch needs to find something CUDA-related to pass this check?

yes, it uses cudart to query the cuda device count: https://github.com/pytorch/pytorch/blob/40ccb7a86d456bc2fb9e45e4b33c774fbf0b3e46/torch/cuda/__init__.py#L125

potentially, it's worth removing this check, because it's a test of the platform not of the package itself. Probably just leaving this check should be fine: `torch.backends.cuda.is_built()`

danpetry:
> find_package(CUDA)

do you want some help with this or are you on it? I don't want to duplicate work but happy to help

h-vetinari:
> do you want some help with this or are you on it? I don't want to duplicate work but happy to help

I'd very much appreciate some help with this. There's just too many `CUDA_*` variables that are (potentially) affected. Let's take this to #333 though to not blow up the thread.

h-vetinari:
Aarch+CUDA timed out after 15h (note time increments per line, as well as some failures)
```
2025-01-29T21:46:49.3943206Z .......s..............ssssss............................................ [ 94%]
2025-01-29T22:18:49.2460065Z ..........FFFFFFF....................................................... [ 95%]
2025-01-29T22:49:07.7088322Z .............s..............................................s........... [ 96%]
2025-01-29T23:28:30.4877456Z ..........ss.ssss......s........................................sss..... [ 96%]
2025-01-29T23:58:21.5877671Z sss...s..sF....................s........................................ [ 97%]
2025-01-30T00:41:33.9959881Z ................................sss................s.....s...s......s... [ 98%]
2025-01-30T00:55:01.2081636Z .....ssssss.......s.s..............ss....................s.............s [ 99%]
2025-01-30T01:15:21.6288586Z ................sss......................s.............................. [ 99%]
2025-01-30T01:17:10.5264273Z ##[error]The operation was canceled.
2025-01-30T01:17:10.5555416Z Post job cleanup.
```
Still, I'm going to merge this for now to get some new builds with the CMake fixes out. I've discussed with @danpetry that it would be good to have some recent builds downloadable for debugging the CUDA part of #333, and so this will help in that respect as well. For that, I'm skipping the CUDA failures mentioned [above](https://github.com/conda-forge/pytorch-cpu-feedstock/pull/318#issuecomment-2615646545), but I'll have yet another PR to try fixing this more properly (+ a bunch of other accumulated stuff).

h-vetinari (approved):
Thanks for the patience here!

h-vetinari:
> yes, it uses cudart to query the cuda device count: https://github.com/pytorch/pytorch/blob/40ccb7a86d456bc2fb9e45e4b33c774fbf0b3e46/torch/cuda/__init__.py#L125
> 
> potentially, it's worth removing this check, because it's a test of the platform not of the package itself. Probably just leaving this check should be fine: `torch.backends.cuda.is_built()` Having said that, it does let the tests fail early if there's no cuda support on the platform, which is necessary to run the tests. I'm surprised it's failing since this is running on a machine with a GPU..?

Well, unfortunately this fails even with the full CUDA toolchain (incl. cudart) present. I don't know how these tests are passing for Anaconda, or what's different here, but this is not working unfortunately.

h-vetinari:
WTH, why does the test not find the GPUs anymore? (this is on MKL; meaning it passed `torch.cuda.is_available()` to get to the python tests)
```
=========================== short test summary info ============================
FAILED [4.2186s] test/test_nn.py::TestNNDeviceTypeCUDA::test_cross_entropy_loss_2d_out_of_bounds_class_index_cuda_float16 - AssertionError: 'CUDA error: device-side assert triggered' not found in 'PYTORCH_API_USAGE torch.python.import\nPYTORCH_API_USAGE c10d.python.import\nPYTORCH_API_USAGE aten.init.cuda\nE\n======================================================================\nERROR: test_cross_entropy_loss_2d_out_of_bounds_class_index (__main__.TestThatContainsCUDAAssert)\n----------------------------------------------------------------------\nTraceback (most recent call last):\n  File "<string>", line 16, in test_cross_entropy_loss_2d_out_of_bounds_class_index\n  File "$PREFIX/lib/python3.10/site-packages/torch/cuda/__init__.py", line 319, in _lazy_init\n    torch._C._cuda_init()\nRuntimeError: No CUDA GPUs are available\n\n----------------------------------------------------------------------\nRan 1 test in 0.006s\n\nFAILED (errors=1)\n'

    python test/test_nn.py TestNNDeviceTypeCUDA.test_cross_entropy_loss_2d_out_of_bounds_class_index_cuda_float16

FAILED [3.2179s] test/test_nn.py::TestNNDeviceTypeCUDA::test_cross_entropy_loss_2d_out_of_bounds_class_index_cuda_float32 - AssertionError: 'CUDA error: device-side assert triggered' not found in 'PYTORCH_API_USAGE torch.python.import\nPYTORCH_API_USAGE c10d.python.import\nPYTORCH_API_USAGE aten.init.cuda\nE\n======================================================================\nERROR: test_cross_entropy_loss_2d_out_of_bounds_class_index (__main__.TestThatContainsCUDAAssert)\n----------------------------------------------------------------------\nTraceback (most recent call last):\n  File "<string>", line 16, in test_cross_entropy_loss_2d_out_of_bounds_class_index\n  File "$PREFIX/lib/python3.10/site-packages/torch/cuda/__init__.py", line 319, in _lazy_init\n    torch._C._cuda_init()\nRuntimeError: No CUDA GPUs are available\n\n----------------------------------------------------------------------\nRan 1 test in 0.004s\n\nFAILED (errors=1)\n'

    python test/test_nn.py TestNNDeviceTypeCUDA.test_cross_entropy_loss_2d_out_of_bounds_class_index_cuda_float32

FAILED [1.8866s] test/test_torch.py::TestTorchDeviceTypeCUDA::test_cublas_config_nondeterministic_alert_cuda - AssertionError: Subprocess exception while attempting to run function "mm" with config "garbage":
  File "<string>", line 10, in <module>
  File "$PREFIX/lib/python3.10/site-packages/torch/cuda/__init__.py", line 319, in _lazy_init
    torch._C._cuda_init()
RuntimeError: No CUDA GPUs are available


    python test/test_torch.py TestTorchDeviceTypeCUDA.test_cublas_config_nondeterministic_alert_cuda

FAILED [5.6304s] test/inductor/test_torchinductor.py::TritonCodeGenTests::test_indirect_device_assert - AssertionError: False is not true : first_arg, 2, True, False

    python test/inductor/test_torchinductor.py TritonCodeGenTests.test_indirect_device_assert

= 4 failed, 14519 passed, 2657 skipped, 91 xfailed, 143276 warnings in 4387.90s (1:13:07) =
```

Anyway, I went back to the state before this PR in 4f1b584e5541baac92971c1a7d681202dfecaacc.

h-vetinari:
Sigh, we really went 4/4 with the CUDA builds failing for various reasons; it looks like the `test_torchinductor` tests cause a massive runtime regression in emulation of linux+aarch, where builds went from ~11h to >15h. I'm going to skip them there. Finally, it seems the torch.compile tests also rely on dynamo, which doesn't have 3.13 support yet.

PR #322, opened 2025-01-15
2.4.x backports

mgorny:
Checklist
* [x] Used a [personal fork of the feedstock to propose changes](https://conda-forge.org/docs/maintainer/updating_pkgs.html#forking-and-pull-requests)
* [x] Bumped the build number (if the version is unchanged)
* [ ] Reset the build number to `0` (if the version changed)
* [x] [Re-rendered]( https://conda-forge.org/docs/maintainer/updating_pkgs.html#rerendering-feedstocks ) with the latest `conda-smithy` (Use the phrase <code>@<space/>conda-forge-admin, please rerender</code> in a comment in this PR for automated rerendering)
* [x] Ensured the license file is being packaged.

Backported the fixes to the two bugs reported against 2.4.x. I haven't rerendered, because this forces CUDA 12.6, and I wasn't able to get it to work — the most obvious changes result in build error:

```
  CMake Error at cmake/public/cuda.cmake:70 (message):
    Failed to find nvToolsExt
  Call Stack (most recent call first):
    cmake/Dependencies.cmake:43 (include)
    CMakeLists.txt:857 (include)
```

and I'm not sure if it's really worth putting more effort into it at this point.

PR #326, opened 2025-01-18
pytorch v2.6.0

h-vetinari:
Build the release candidates

Linux CI cancelled until builds for #322 are live


h-vetinari:
This looks better than expected so far. Still have to double check dependency changes. Happy if someone could do that (even if it's just noting which bounds changed relative to the current recipe)

recipe/meta.yaml
```diff
@@ -305,27 +288,28 @@ outputs:
         - typing_extensions
         - {{ pin_subpackage('libtorch', exact=True) }}
       run:
+        - {{ pin_subpackage('libtorch', exact=True) }}
```
hmaarrfk, line 291: The old syntax was to workaround the fact that for the non megabuilds each libtorch may have different hashes making it incompatible.

hmaarrfk: Or maybe something like that. Maybe you have addressed the core problem and this is the better way.

h-vetinari: Even for the megabuild, libtorch has a unique hash, because there's only one. It's pytorch itself that gets different hashes due to the different python versions

isuruf: > Even for the megabuild, libtorch has a unique hash

Not true. See https://anaconda.org/conda-forge/libtorch/files

h-vetinari: Obviously cuda and non-CUDA create different libtorch hashes, but _within_ a megabuild it's unique, which is what matters for pinning it in pytorch.

Unless the idea was to allow mixing CUDA-enabled pytorch with non-CUDA libtorch, but I don't see the sense in that.

h-vetinari: It's more interesting for the blas_impl - `pytorch` could theoretically be independent of that (if all the blas calls go through libtorch), but also there we're already creating different `pytorch` hashes there due to the `{{ pin_subpackage("libtorch", exact=True) }}` in the host dependencies, so AFAICT we're not materially changing the various installations here, just making it impossible to install untested/unsupported combinations

isuruf: For non-megabuilds, the idea was to allow libtorch from any of the builds (with the same features except python version) to work with any pytorch build. This way, you don't have to download a different `libtorch` for different python version. Note that this is for non-megabuilds only i.e. osx where we don't have CUDA builds.

h-vetinari: Yeah, that makes sense! Let me try to reflect that in the run-deps.

h-vetinari:
Sigh, since when is conda-build applying patches through `git` rather than through `patch`? The former is stricter than the latter, and doesn't work in some situations (like applying a patch in one of the submodules):
```
Applying patch: /Users/runner/work/1/s/recipe/patches_submodules/0001-Fix-bazel-linux-aarch64-gcc13-workflow-and-resolve-a.patch
Applying: Fix `bazel-linux-aarch64-gcc13` workflow and resolve accompanying build errors.
error: sha1 information is lacking or useless (third_party/XNNPACK/src/reference/unary-elementwise.cc).
error: could not build fake ancestor
```

danpetry:
on osx-64
```
FAILED [0.0518s] test/test_nn.py::TestNN::test_batchnorm_nhwc_cpu - AssertionError: Tensor-likes are not close!

Mismatched elements: 1 / 8 (12.5%)
Greatest absolute difference: 1.430511474609375e-05 at index (0,) (up to 1e-05 allowed)
Greatest relative difference: 4.616721980710281e-06 at index (0,) (up to 1.3e-06 allowed)
```
skip?

h-vetinari:
Yeah, this minor accuracy violation indeed sounds skippable, but I've deprioritised this PR until we get the windows builds for 2.5 fixed (and ideally your #318) merged as well.

danpetry:
ok, good to know

danpetry:
worth pointing out that as of 6 days time, pypi will have an up to date pytorch package whereas conda won't. Will have a look at that other PR

h-vetinari:
> worth pointing out that as of 6 days time, pypi will have an up to date pytorch package whereas conda won't.

Are you talking about rc's, or are we not looking at the same [index](https://pypi.org/project/torch/#history)? 2.6.0 GA hasn't been published AFAICT. Or are you saying that 2.6.0 will be released in 6 days?

In any case, this is no reason to rush. We didn't have windows packages for years, and I'm more concerned about fixing them, than lagging behind the PyPI release a bit (and we've often lagged for months in the past; this has gotten much better with the open-gpu server, but it still happens; 2.5.0 was released Oct 18th last year, we had first builds on Nov 3rd).

danpetry:
> are you saying that 2.6.0 will be released in 6 days

yes

> I'm more concerned about fixing them, than lagging behind the PyPI release a bit

100%

recipe/meta.yaml
```diff
@@ -388,7 +384,7 @@ outputs:
         # tools/ is needed to optimise test run
         # as of pytorch=2.0.0, there is a bug when trying to run tests without the tools
         - tools
-        #- .ci/pytorch/smoke_test/smoke_test.py
+        - .ci/pytorch/smoke_test/smoke_test.py
```
h-vetinari, line 387: @danpetry, it seems this file is not part of the tarball:
```
RuntimeError: Did not find any source_files for test with pattern .ci/pytorch/smoke_test/smoke_test.py
```

h-vetinari:
Tried enabling the torchinductor tests on CPU builds (the upstream test module does have skips if GPUs are missing), but running into the following on OSX:
```
----------------------------- Captured stderr call -----------------------------
OMP: Error #179: Function pthread_key_create failed:
OMP: System error #35: Resource temporarily unavailable
```
Also the test suite seems to be segfaulting a lot, but somehow pytest survives until the end to show the results.

h-vetinari:
Win+CUDA was failing with:
```
E              Creating library C:/Users/RUNNER~1.CIR/AppData/Local/Temp/tmp2jc5ie40/gp/cgpa5q4n3ccplmfmz7gkg7vnulupieu2kqkng2abexlk7g6kae3f.lib and object C:/Users/RUNNER~1.CIR/AppData/Local/Temp/tmp2jc5ie40/gp/cgpa5q4n3ccplmfmz7gkg7vnulupieu2kqkng2abexlk7g6kae3f.exp
E           cgpa5q4n3ccplmfmz7gkg7vnulupieu2kqkng2abexlk7g6kae3f.obj : error LNK2019: unresolved external symbol aoti_torch_check referenced in function kernel
E           C:\Users\RUNNER~1.CIR\AppData\Local\Temp\tmp2jc5ie40\gp\cgpa5q4n3ccplmfmz7gkg7vnulupieu2kqkng2abexlk7g6kae3f.pyd : fatal error LNK1120: 1 unresolved externals
```
This was handled in https://github.com/pytorch/pytorch/pull/140030

recipe/meta.yaml
danpetry: we could also do something like [this](https://github.com/AnacondaRecipes/pytorch-feedstock/blob/4f8a8bbf9cbcf41605ccddceaa1f4151ad609de3/recipe/meta.yaml#L58) to remove the need to manually update. Maybe later.

recipe/patches/0020-make-library-name-in-test_mutable_custom_op_fixed_la.patch
h-vetinari, line 1: @danpetry, so I had taken your [suggestion](https://github.com/conda-forge/pytorch-cpu-feedstock/pull/318#discussion_r1932908123) and added another fix in 138456c736799758ca263ab2448d5c7d5168fa21, but it still fails with
```
E   ModuleNotFoundError: No module named 'torch.ops.mylib_140714837156224'
```
Perhaps I'm overlooking something, if so, happy to experiment more on this. In the meantime, I've re-skipped those tests (5ee95f4bc7f954bef665ee3daecb4ae1c18903bf), and so I'm dropping the patch too.

Also worth recalling the original failure that lead to us looking at this:
```
>       self.m: Optional[Any] = torch._C._dispatch_library(
            kind, ns, dispatch_key, filename, lineno
        )
E       RuntimeError: Only a single TORCH_LIBRARY can be used to register the namespace mylib; please put all of your definitions in a single TORCH_LIBRARY block.  If you were trying to specify implementations, consider using TORCH_LIBRARY_IMPL (which can be duplicated).  If you really intended to define operators for a single namespace in a distributed way, you can use TORCH_LIBRARY_FRAGMENT to explicitly indicate this.  Previous registration of TORCH_LIBRARY was registered at C:\Users\runnerx.cirun-8e2044a\AppData\Local\torch_extensions\torch_extensions\Cache\py312_cu126\mylib\main.cpp:23; latest registration was registered at /dev/null:137
```

danpetry: yes, saw that, sounds fine.

danpetry: a future pr could involve enabling and fixing torchinductor

h-vetinari: The torchinductor tests _do_ run, but only for one python version (due to the long test runtime; adding that module makes the overall test time go from ~7-8min to around 35-40).

recipe/meta.yaml
h-vetinari: ah, should have checked the anaconda recipe. Fine for me to use the same approach here!

h-vetinari:
The smoketest isn't working on linux-aarch:
```
+ python ./smoke_test.py --package torchonly
torch: 2.6.0
ATen/Parallel:
	at::get_num_threads() : 4
	at::get_num_interop_threads() : 8
OpenMP 201511 (a.k.a. OpenMP 4.5)
	omp_get_max_threads() : 4
MKL not found
Intel(R) MKL-DNN v3.5.3 (Git Hash c0bd3b8cc25ee5fd5d8c7f273f7dabd6913d11f3)
std::thread::hardware_concurrency() : 8
Environment variables:
	OMP_NUM_THREADS : 4
	MKL_NUM_THREADS : [not set]
ATen parallel backend: OpenMP

Testing smoke_test_conv2d
/home/conda/feedstock_root/build_artifacts/libtorch_1738637456711/work/third_party/ideep/mkl-dnn/src/cpu/aarch64/xbyak_aarch64/src/util_impl_linux.h, 444: Can't open MIDR_EL1 sysfs entry
Testing smoke_test_linalg on cpu
Testing smoke_test_compile for cpu and torch.float16
/home/conda/feedstock_root/build_artifacts/libtorch_1738637456711/test_tmp/run_test.sh: line 21: 35326 Segmentation fault      (core dumped) python ./smoke_test.py --package torchonly
```

Could be an issue with emulation of course. I'll skip it for now.

h-vetinari:
OK, so my gambit with https://github.com/conda-forge/triton-feedstock/commit/b379a984281d8b8529680ad2feafbd0a996a089f didn't work. Unless there's another way to do this that I'm not aware of, it seems we'll either need to publish `libllvm20` to `main`, or ask every user of CUDA-enabled pytorch to manually add `-c conda-forge/label/llvm_rc` (which is a very unappealing option).

h-vetinari:
Smoketest is not looking good on linux-64 + CUDA:
```
Testing smoke_test_conv2d
Testing smoke_test_conv2d with cuda
  File "$TEST_SRC/./smoke_test.py", line 388, in <module>
    main()
    ~~~~^^
  File "$TEST_SRC/./smoke_test.py", line 373, in main
    smoke_test_conv2d()
    ~~~~~~~~~~~~~~~~~^^
  File "$TEST_SRC/./smoke_test.py", line 233, in smoke_test_conv2d
    conv = nn.Conv2d(3, 3, 3).cuda()
  File "$PREFIX/lib/python3.13/site-packages/torch/nn/modules/module.py", line 1053, in cuda
    return self._apply(lambda t: t.cuda(device))
           ~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "$PREFIX/lib/python3.13/site-packages/torch/nn/modules/module.py", line 930, in _apply
    param_applied = fn(param)
  File "$PREFIX/lib/python3.13/site-packages/torch/nn/modules/module.py", line 1053, in <lambda>
    return self._apply(lambda t: t.cuda(device))
                                 ~~~~~~^^^^^^^^
  File "$PREFIX/lib/python3.13/site-packages/torch/cuda/__init__.py", line 319, in _lazy_init
    torch._C._cuda_init()
    ~~~~~~~~~~~~~~~~~~~^^
RuntimeError: No CUDA GPUs are available
```

That's despite having a full CUDA toolchain in the environment and a GPU on the machine... 🤔

h-vetinari:
Based on the [comparison](https://github.com/conda-forge/pytorch-cpu-feedstock/pull/339#issuecomment-2633042211) with the last build where linux-64 + CUDA passed, I'm going to try reverting https://github.com/conda-forge/pytorch-cpu-feedstock/commit/9fcb3a7850e987c6560fe186343f20b99196fd37. The only other _potentially_ relevant difference I can spot is e1f50acc5ad48ba53be14386e1e8425852fa4a16, but I fail to see how that would affect anything.

h-vetinari:
OK, the smoke test is not adding value here but just running into problems that aren't relevant:
```
    torch._C._cuda_init()
2025-02-05T16:21:34.6159587Z RuntimeError: No CUDA GPUs are available
```
In line with @hmaarrfk's [comment](https://github.com/conda-forge/pytorch-cpu-feedstock/pull/339#issuecomment-2636951863), I'm removing it.

hmaarrfk:
Thanks the middle ground is hard. 

I don't dismiss the tremendous effort you put into this. I'm just worried about the day you decide you can't put in the effort anymore.

mgorny:
I'm thinking we need to reconsider which tests to run. I kinda chose them semi-randomly, and they definitely test something meaningful, but the number of failures is getting absurd. I mean, if the only result of running the test suite is adding more skips, then I dare say it's not a very meaningful suite.

h-vetinari:
> if the only result of running the test suite is adding more skips, then I dare say it's not a very meaningful suite.

not sure I get what you're saying, aside from skipping the torchinductor tests on osx again (which were only enabled for CPU-builds in 2182f8145d0cd1737f5a365a74ee7a3aedcdb017 / https://github.com/conda-forge/pytorch-cpu-feedstock/pull/340), this PR is adding one(!) new test skip, out of a total of >7500 tests. I think it's definitely serving its purpose for now, though of course I'm open to rethink the set of modules we use. I'm also planning to reduce the skips to the bare minimum (see work started in #352 & #353).

mgorny:
I was just saying in general, it feels like we're skipping too many tests.

danpetry (commented):
I had a look over and can't see anything I disagree with.
Agree that the tests and also the patches are getting quite extensive. Suggest that we work to reduce that in future PRs perhaps.

h-vetinari:
> I was just saying in general, it feels like we're skipping too many tests.

I don't think it's that bad TBH. We're skipping 35 tests (out of 8000+), across all architectures, and the vast majority of those skips are specific to one or two variants in our matrix. That's roughly 0.1-0.2% of tests getting skipped on a given platform.

> Agree that the tests and also the patches are getting quite extensive. Suggest that we work to reduce that in future PRs perhaps.

💯 I've started an effort to reduce the skips (see #353 & #353). Reducing the patches would need a dedicated effort on the part of the respective authors to upstream their fix. I'll try upstreaming the `find_package(CUDA)` patch, but it's a non-trivial effort, and even comparatively simple ones like https://github.com/pytorch/pytorch/pull/137084 are still open months later. OTOH, I did structure my include-path manipulation patches in a way that I could at least upstream https://github.com/pytorch/pytorch/pull/145480.

Overall, both the patches and the skips are aiming to be the minimal set necessary to successfully build and test pytorch. I'd love to see them reduced as fast as possible, but for now I don't have a better alternative.

danpetry:
For now, I haven't reviewed in-depth enough to be able to say "approval", I think - could do that tomorrow if it's needed.
Otherwise: ok to proceed from me.

h-vetinari:
Oh boy, nevermind. Something is very wrong on linux+CUDA:
```
= 191 failed, 14397 passed, 2763 skipped, 91 xfailed, 143695 warnings in 4444.09s (1:14:04) =
```

It seems that many (56) are due to `RuntimeError: cuDNN error: CUDNN_STATUS_INTERNAL_ERROR_HOST_ALLOCATION_FAILED`, so basically: running out of memory. Even more (130) are due to `RuntimeError: GET was unable to find an engine to execute this computation`, which probably has the same underlying reason.

h-vetinari:
There's another failure that looks like it could have been due to the OOMs, but surprisingly, it fails in exactly the same way on MKL/openblas, and with exactly the same (catastrophic) accuracy violation:

```
____________________ TestNN.test_Conv3d_1x1x1_no_bias_cuda _____________________
[...]
E                   AssertionError: Tensor-likes are not close!
E                   
E                   Mismatched elements: 6 / 6 (100.0%)
E                   Greatest absolute difference: 18.49855402601494 at index (1, 1, 0, 0, 0) (up to 0.0002 allowed)
E                   Greatest relative difference: 211.43751630108665 at index (0, 0, 0, 0, 0) (up to 0 allowed)
```

<details>
<summary>stacktrace</summary>

```
____________________ TestNN.test_Conv3d_1x1x1_no_bias_cuda _____________________
[gw1] linux -- Python 3.10.16 $PREFIX/bin/python

self = <test_nn.TestNN testMethod=test_Conv3d_1x1x1_no_bias_cuda>
test = <torch.testing._internal.common_nn.NewModuleTest object at 0x7fc5465f89a0>
kwargs = {}

    def with_tf32_off(self, test=test, kwargs=kwargs):
        with tf32_off():
>           test.test_cuda(self, **kwargs)

test/test_nn.py:7322: 

self = <torch.testing._internal.common_nn.NewModuleTest object at 0x7fc5465f89a0>
test_case = <test_nn.TestNN testMethod=test_Conv3d_1x1x1_no_bias_cuda>

    def test_cuda(self, test_case):
        if not TEST_CUDA or not self.should_test_cuda:
            raise unittest.SkipTest('Excluded from CUDA tests')
    
        with set_default_dtype(self.default_dtype):
            cpu_input = self._get_input()
    
            type_map = {torch.double: torch.float}
            cpu_input_tuple = cpu_input if isinstance(cpu_input, tuple) else (cpu_input,)
    
            is_any_input_complex = any(isinstance(t, torch.Tensor) and t.dtype.is_complex for t in cpu_input_tuple)
    
            gpu_input_tuple = to_gpu(cpu_input_tuple, type_map=type_map)
    
            cpu_module = self.constructor(*self.constructor_args)
            gpu_module = self.constructor(*self.constructor_args).float().cuda()
            cpu_param = test_case._get_parameters(cpu_module)
            gpu_param = test_case._get_parameters(gpu_module)
            for cpu_p, gpu_p in zip(cpu_param[0], gpu_param[0]):
                gpu_p.data.copy_(cpu_p)
    
            test_case._zero_grad_input(cpu_input_tuple)
            test_case._zero_grad_input(gpu_input_tuple)
            test_case._zero_grad_parameters(cpu_module)
            test_case._zero_grad_parameters(gpu_module)
            cpu_output = test_case._forward(cpu_module, cpu_input_tuple)
            gpu_output = test_case._forward(gpu_module, gpu_input_tuple)
            if getattr(cpu_module, "return_indices", False):
                cpu_output = cpu_output[0]
                gpu_output = gpu_output[0]
            test_case.assertEqual(cpu_output, gpu_output, atol=self.precision, rtol=0, exact_dtype=False)
    
            # Run backwards on CPU and GPU and compare results
            for _ in range(5):
                cpu_gradOutput = cpu_output.clone().normal_()
                gpu_gradOutput = cpu_gradOutput.type_as(gpu_output)
                cpu_gradInput = test_case._backward(cpu_module, cpu_input_tuple, cpu_output, cpu_gradOutput)
                gpu_gradInput = test_case._backward(gpu_module, gpu_input_tuple, gpu_output, gpu_gradOutput)
                test_case.assertEqual(cpu_gradInput, gpu_gradInput, atol=self.precision, rtol=0, exact_dtype=False)
                for cpu_d_p, gpu_d_p in zip(cpu_param[1], gpu_param[1]):
                    test_case.assertEqual(cpu_d_p, gpu_d_p, atol=self.precision, rtol=0)
    
            # Run double-backwards on CPU and GPU and compare results
            if self.check_gradgrad and not self.FIXME_no_cuda_gradgrad_comparison:
                cpu_output = cpu_module(*cpu_input_tuple)
                gpu_output = gpu_module(*gpu_input_tuple)
                if getattr(cpu_module, "return_indices", False):
                    cpu_output = cpu_output[0]
                    gpu_output = gpu_output[0]
    
                cpu_gradOutput = torch.randn_like(cpu_output, requires_grad=True)
                gpu_gradOutput = cpu_gradOutput.type_as(gpu_output).detach()
                gpu_gradOutput.requires_grad = True
    
                cpu_gradInputs = torch.autograd.grad(
                    cpu_output,
                    cpu_input_tuple + tuple(cpu_module.parameters()),
                    cpu_gradOutput,
                    create_graph=True)
                gpu_gradInputs = torch.autograd.grad(
                    gpu_output,
                    gpu_input_tuple + tuple(gpu_module.parameters()),
                    gpu_gradOutput,
                    create_graph=True)
    
                for cpu_d_i, gpu_d_i in zip(cpu_gradInputs, gpu_gradInputs):
>                   test_case.assertEqual(cpu_d_i, gpu_d_i, atol=self.precision, rtol=0, exact_dtype=False)
E                   AssertionError: Tensor-likes are not close!
E                   
E                   Mismatched elements: 6 / 6 (100.0%)
E                   Greatest absolute difference: 18.49855402601494 at index (1, 1, 0, 0, 0) (up to 0.0002 allowed)
E                   Greatest relative difference: 211.43751630108665 at index (0, 0, 0, 0, 0) (up to 0 allowed)
```

</details>

danpetry:
A lot of the skips look like they're there to cope with CI limitations, basically

h-vetinari:
> A lot of the skips look like they're there to cope with CI limitations, basically

Sure, some portion of those are unavoidable, but basically all skips point at an issue that should be fixed in our packaging, or ideally upstream (in the test framework and/or the actual implementation). I kinda prefer having that transparency (and as a reminder that there's more work to be done), than simply removing the test modules in question and declaring it a done deal.

For example, the torchinductor tests you added in #318 were painful to get passing, but they did show that the `torch.compile` feature had essentially be broken in our packaging. I prefer a longer list of skips than a less functional package through ignoring problematic test modules wholesale. Of course, the best is to work through the issues surfaced by the skips, but that needs to happen on a longer timeline IMO, not while we're trying to get the release out.

h-vetinari:
Since the linux64+openblas+CUDA job had already passed for https://github.com/conda-forge/pytorch-cpu-feedstock/pull/326/commits/f14feb5e3e043b2a59fd6776d504ba72f052e0ff (with no essential changes since then), the only thing to do is to remove the `blas_impl == "mkl"` from the selector of this line
https://github.com/conda-forge/pytorch-cpu-feedstock/blob/be203905856551e5a4b50dd03c6be2e9ddef5fc4/recipe/meta.yaml#L502-L503

As such, I'm not planning to do a new push here (will shift the skip when merging). I'm planning to merge this in ~9-12 hours unless there are other comments @conda-forge/pytorch-cpu; I don't doubt that we'll keep iterating, so even post-merge comments won't take long to get picked up.

danpetry:
🎉 thanks a lot for the huge amount of work on this over the last weeks @h-vetinari . This is really great.

PR #328, opened 2025-01-19
Move libtorch bits to PREFIX & enable cuDSS, cuSPARSELt, CUPTI on win; fix duplicate `.pyc` versions

### LINKED ISSUE #319: Enabling CUDA CUPTI on Windows
Currently CUPTI is only used on Linux:

https://github.com/conda-forge/pytorch-cpu-feedstock/blob/39b84c33b7e8c4c5bc81986602cfb1f5b3d39e28/recipe/meta.yaml#L129

However there are [Windows packages for CUPTI]( https://github.com/conda-forge/cuda-cupti-feedstock/blob/2caf1691ed64a9b526771495515ffe4050026d84/recipe/meta.yaml#L81-L93 )

Would be interested to know if there were issues encountered during Windows bringup that need to be addressed

As it would be great if we could supply CUPTI enabled builds to Windows users

### LINKED ISSUE #320: Enabling cuSPARSELt on Windows
Currently cuSPARSELt is only used on Linux:

https://github.com/conda-forge/pytorch-cpu-feedstock/blob/39b84c33b7e8c4c5bc81986602cfb1f5b3d39e28/recipe/meta.yaml#L134

However there are [Windows packages for cuSPARSELt]( https://github.com/conda-forge/cusparselt-feedstock/blob/4413abaf4468c3efe239318afbd149dc4cfe40b6/recipe/meta.yaml#L24-L26 )

Would be interested to know if there were issues encountered during Windows bringup that need to be addressed

As it would be great if we could supply cuSPARSELt enabled builds to Windows users

### LINKED ISSUE #321: Enabling cuDSS on Windows
Currently cuDSS is only used on Linux:

https://github.com/conda-forge/pytorch-cpu-feedstock/blob/39b84c33b7e8c4c5bc81986602cfb1f5b3d39e28/recipe/meta.yaml#L136

However there are [Windows packages for cuDSS]( https://github.com/conda-forge/libcudss-feedstock/blob/f4cb6db821a10b3a665748ea96d2c492fce923a6/recipe/meta.yaml#L100-L102 )

Would be interested to know if there were issues encountered during Windows bringup that need to be addressed

As it would be great if we could supply cuDSS enabled builds to Windows users

### LINKED ISSUE #327: Windows C extension compilation issues
### Solution to issue cannot be found in the documentation.

- [x] I checked the documentation.

### Issue

https://github.com/conda-forge/sam-2-feedstock/pull/16

Shows some errors in inspecting windows linked libraries.

I feel like it is similar to the challenges we faced moving the shared libraries around for Linux. 

### Installed packages

```shell
PyTorch CUDA 12.6 2.5.1
```

### Environment info

```shell
Windows azure CI with CUDA build environment.
```

### LINKED ISSUE #329: BUG: Duplicated `.pyc` files for different python version
I was looking at some logs in more detail, and noticed that for all versions `py!=313`, we're getting a full duplication of the `.pyc` files across the entire `$SP_DIR/torch` tree, e.g.
```
?rw-rw-r-- 0/0       6514 2025-01-20 13:16:20 lib/python3.9/site-packages/torch/mps/__pycache__/__init__.cpython-313.pyc 
?rw-rw-r-- 0/0       5519 2025-01-20 13:16:14 lib/python3.9/site-packages/torch/mps/__pycache__/__init__.cpython-39.pyc 
?rw-rw-r-- 0/0       3128 2025-01-20 13:16:20 lib/python3.9/site-packages/torch/mps/__pycache__/event.cpython-313.pyc 
?rw-rw-r-- 0/0       2583 2025-01-20 13:16:14 lib/python3.9/site-packages/torch/mps/__pycache__/event.cpython-39.pyc 
?rw-rw-r-- 0/0       2914 2025-01-20 13:16:20 lib/python3.9/site-packages/torch/mps/__pycache__/profiler.cpython-313.pyc 
?rw-rw-r-- 0/0       2900 2025-01-20 13:16:14 lib/python3.9/site-packages/torch/mps/__pycache__/profiler.cpython-39.pyc 
                     ^^^^
                     !!!!
```
As we can see from the column marked with `^`, this actually has a pretty substantial size impact (because even before the duplication, there are about 2000 `.pyc` files in `pytorch`).

On windows we have the same problem, namely that we similarly [see](https://conda-metadata-app.streamlit.app/?q=conda-forge%2Fwin-64%2Fpytorch-2.5.1-cuda126_mkl_py39_h5dc7ac5_309.conda&with_broken=true)
```
lib/python3.9/site-packages/torch/mps/__pycache__/__init__.cpython-313.pyc
lib/python3.9/site-packages/torch/mps/__pycache__/__init__.cpython-39.pyc
lib/python3.9/site-packages/torch/mps/__pycache__/event.cpython-313.pyc
lib/python3.9/site-packages/torch/mps/__pycache__/event.cpython-39.pyc
lib/python3.9/site-packages/torch/mps/__pycache__/profiler.cpython-313.pyc
lib/python3.9/site-packages/torch/mps/__pycache__/profiler.cpython-39.pyc
```
again for all versions but `py!=313`.

mgorny:
Checklist

Fixes #319 
Fixes #320 
Fixes #321
Fixes #327 
Fixes #329


h-vetinari:
Would also be good to address #327 in this PR.

hmaarrfk:
(draft until all builds are enabled again)

h-vetinari:
So the extra CUDA deps on windows seem to work fine 🥳

mgorny:
Ok, I've decided to instead start by adding error handling to `.bat`, to see which of the existing commands fail.

By the way, any chance you could give me the power to run CI myself?

h-vetinari:
> By the way, any chance you could give me the power to run CI myself?

You already should have AFAIU, and I don't have the power to change it further. This should be a matter of https://github.com/Quansight/open-gpu-server/pull/52 plus a sync of the setup to https://github.com/conda-forge/.cirun (which happened), so I'm really not sure where this goes wrong. Pinging @jaimergp & @aktech

One conceivable angle is that the lack of authorization for the windows agents prevents you from starting any job at all (though IIRC you already weren't able to start things before we merged windows support). For that, you could make a PR like https://github.com/conda-forge/.cirun/pull/27

recipe/bld.bat
```diff
@@ -170,46 +173,52 @@ if errorlevel 1 exit /b 1
...
-    pushd torch-*
+    pushd torch-* || goto :error
 
     @REM Move the binaries into the packages site-package directory
-    robocopy /NP /NFL /NDL /NJH /E torch\bin %SP_DIR%\torch\bin\
-    robocopy /NP /NFL /NDL /NJH /E torch\lib %SP_DIR%\torch\lib\
-    robocopy /NP /NFL /NDL /NJH /E torch\share %SP_DIR%\torch\share\
+    robocopy /NP /NFL /NDL /NJH /E torch\bin %SP_DIR%\torch\bin\ || goto :error
+    robocopy /NP /NFL /NDL /NJH /E torch\lib %SP_DIR%\torch\lib\ || goto :error
+    robocopy /NP /NFL /NDL /NJH /E torch\share %SP_DIR%\torch\share\ || goto :error
```
h-vetinari, line 175: You need to be careful with robocopy, which has an exit code of 1 if files were successfully copied (I [learned](https://github.com/conda-forge/clangdev-feedstock/blob/d5a5e4aff22133083f5c8ab87a100a2652f3d935/recipe/bld.bat#L7-L8) the hard way too).

h-vetinari (commented):
I'd prefer to add
```batch
if %ERRORLEVEL% neq 0 exit 1
```
to calls that can fail, rather than adding `|| goto error` everywhere. There's no functional difference AFAIK, and although both styles have their pros & cons, the goto-variant is quite rare in conda-forge (a quick search [shows](https://github.com/search?q=org%3Aconda-forge+%2Fgoto+%3Aerror%2F+path%3A*.bat&type=code) 170 feedstocks using goto, while ~1.6 feedstocks [use](https://github.com/search?q=org%3Aconda-forge+%2Fif+%5C%25%3Ferrorlevel%2F+path%3A*.bat&type=code) an errorlevel variant).

aktech:
@mgorny You need to add your name here: https://github.com/conda-forge/.cirun/blob/218150c90899c7dfeb771d0bd9b176ca191848f3/.access.yml#L89 like this PR: https://github.com/conda-forge/.cirun/pull/27 and this is subject to @wolfv's approval I believe as they (prefix.dev) are the ones sponsoring the windows machine.

and you need to be in any of [these roles](https://github.com/conda-forge/.cirun/blob/218150c90899c7dfeb771d0bd9b176ca191848f3/.access.yml#L37) for this repo to be able to trigger runs that involves GPU CI runs. cc @jaimergp

h-vetinari (dismissed):
We're going to need existence tests for the DLLs to be under `%LIBRARY_BIN%`.

The tests are already there, and just need to be modified, e.g.
https://github.com/conda-forge/pytorch-cpu-feedstock/blob/39b84c33b7e8c4c5bc81986602cfb1f5b3d39e28/recipe/meta.yaml#L199
(similar below)

h-vetinari:
Thanks for the quick response @aktech!

> and you need to be in any of [these roles](https://github.com/conda-forge/.cirun/blob/218150c90899c7dfeb771d0bd9b176ca191848f3/.access.yml#L37) for this repo to be able to trigger workflows that involves GPU CI runs.

Ah, perhaps it's just a question of making @mgorny a maintainer here? That's a hypothesis we can test easily (given your involvement here, I'm assuming you have nothing against this @mgorny) 😉

mgorny:
Ok, I'll replace `|| goto :error` with `if errorlevel` checks, then push and take a break.

h-vetinari:
> We'll find out on your next push here!

I was talking about the linux builds. For the windows agents, we'll definitely still need to wait for https://github.com/conda-forge/.cirun/pull/32

h-vetinari:
> Ok, I'll replace `|| goto :error` with `if errorlevel` checks

You're probably aware, but just for completeness (and because batch is an insane language): `if errorlevel` is a very different construct from `if %errorlevel%`. In particular, the more "advanced" relational operators like `NEQ` are not [supported](https://stackoverflow.com/a/74148543); more [details](https://devblogs.microsoft.com/oldnewthing/20080926-00/?p=20743)

mgorny:
Ah, okay, sorry. Either case, I really need to take a break now, so maybe it'll be done when I get back.

h-vetinari:
> Ah, okay, sorry. Either case, I really need to take a break now, so maybe it'll be done when I get back.

No worries. Perhaps until then the PR is merged. ;-)

When you do get back, please squash https://github.com/conda-forge/pytorch-cpu-feedstock/pull/328/commits/f9fe7bfa4b1351c1c95ceaaf3f3f0f9ffe3655d4 & https://github.com/conda-forge/pytorch-cpu-feedstock/pull/328/commits/ad726bc45463c651208092aa2688d76a4c34105c to avoid the back-and-forth.

mgorny:
Okay, let's hope this is correct.

Could someone trigger the Windows build? :-)

recipe/meta.yaml
```diff
@@ -15,7 +15,7 @@
 # https://github.com/pytorch/pytorch/blob/main/RELEASE.md#triton-dependency-for-the-release
 {% set triton = "3.1.0" %}
 
-# TODO Temporary pin, remove
+# TODO Temporary pin, remove me
 {% set mkl = "<2025" %}
```
hmaarrfk, line 19: where was this hold over from?

mgorny: I think it was added while @Tobias-Fischer was trying to figure out why Windows builds were failing earlier.

hmaarrfk: hmm. well we probably need to revert ;)

recipe/bld.bat
```diff
@@ -221,14 +221,13 @@ if "%PKG_NAME%" == "libtorch" (
...
 
-    @REM Delete all files from the lib directory that do not start with torch_python
-    for %%f in (%SP_DIR%\torch\lib\*) do (
-        set "FILENAME=%%~nf"
-        if "!FILENAME:~0,12!" neq "torch_python" (
-            del %%f
-            if %ERRORLEVEL% neq 0 exit 1
-        )
-    )
+    @REM Move libtorch_python and remove the lib directory afterwards.
```
h-vetinari, line 224: The python-specific lib doesn't have to go into `%LIBRARY_BIN%` necessarily. We should match what the unix build does there.

recipe/meta.yaml
h-vetinari: Actually, I don't mind this for now. MKL 2025 has a lot of problems (or we have problems with it) - it's still not part of our blas infrastructure due to https://github.com/conda-forge/intel_repack-feedstock/issues/83

So as a measure to reduce the opportunities for the solver to get lost trying to (wrongly) update to mkl 2025, this is beneficial

recipe/bld.bat
```diff
@@ -175,46 +172,49 @@ if errorlevel 1 exit /b 1
...
+    if %ERRORLEVEL% neq 0 exit 1
 
     @REM Move the binaries into the packages site-package directory
-    robocopy /NP /NFL /NDL /NJH /E torch\bin %SP_DIR%\torch\bin\
-    robocopy /NP /NFL /NDL /NJH /E torch\lib %SP_DIR%\torch\lib\
-    robocopy /NP /NFL /NDL /NJH /E torch\share %SP_DIR%\torch\share\
+    robocopy /NP /NFL /NDL /NJH /E torch\bin\ %LIBRARY_BIN%\
+    robocopy /NP /NFL /NDL /NJH /E torch\lib\ %LIBRARY_BIN%\ *.dll
+    robocopy /NP /NFL /NDL /NJH /E torch\lib\ %LIBRARY_LIB%\ *.lib
+    robocopy /NP /NFL /NDL /NJH /E torch\share %LIBRARY_PREFIX%\
```
h-vetinari, line 191: I think the syntax here doesn't work as for `cp`, and will need something like
```suggestion
    robocopy /NP /NFL /NDL /NJH /E torch\share\ %LIBRARY_PREFIX%\share
```

h-vetinari: We should also add a smoketest for a given file that we expect to be under `%LIBRARY_PREFIX%\share`.

recipe/bld.bat
mgorny: Are we talking of `.lib` or both `.lib` and `.dll`? I think only the former would be needed for linking, and that one should indeed be tiny.

mgorny: `.lib` is 220k, `.dll` is 14M.

h-vetinari: That sounds good to me, but it probably doesn't work with a simple copy then. I haven't looked into how to change the path that the `.lib` expects the `.dll` in, but I assume it's possible

mgorny: But does it contain a path at all? At least according to a quick look, it seems to contain just the `torch_python.dll` filename, and I can't grep any `site-packages` or `python3` in it either.

h-vetinari: OK, fair enough. Let's try just copying the `.lib`

mgorny:
```
2025-01-21T17:59:43.7527514Z import: 'torch'
2025-01-21T17:59:43.8354491Z Traceback (most recent call last):
2025-01-21T17:59:43.8355076Z   File "C:\bld\libtorch_1737467871481\test_tmp\run_test.py", line 2, in <module>
2025-01-21T17:59:43.8355584Z     import torch
2025-01-21T17:59:43.8356093Z   File "C:\bld\libtorch_1737467871481\_test_env\lib\site-packages\torch\__init__.py", line 367, in <module>
2025-01-21T17:59:43.8357656Z     from torch._C import *  # noqa: F403
2025-01-21T17:59:43.8358371Z ImportError: DLL load failed while importing _C: The specified module could not be found.
```

Also, apparently some copying failed:

```
2025-01-21T17:52:25.5443176Z 2025/01/21 17:52:25 ERROR 123 (0x0000007B) Accessing Source Directory C:\bld\libtorch_1737467871481\work\torch\lib\torch_python.dll\
2025-01-21T17:52:25.5443825Z The filename, directory name, or volume label syntax is incorrect.
2025-01-21T17:52:25.5444090Z 
2025-01-21T17:52:25.5542861Z 
2025-01-21T17:52:25.5546315Z 2025/01/21 17:52:25 ERROR 123 (0x0000007B) Accessing Source Directory C:\bld\libtorch_1737467871481\work\torch\lib\torch_python.lib\
2025-01-21T17:52:25.5547390Z The filename, directory name, or volume label syntax is incorrect.
2025-01-21T17:52:25.5547856Z 
2025-01-21T17:52:25.6443250Z 
2025-01-21T17:52:25.6447530Z 2025/01/21 17:52:25 ERROR 2 (0x00000002) Accessing Source Directory C:\bld\libtorch_1737467871481\_h_env\Library\lib\torch_python.lib\
2025-01-21T17:52:25.6448524Z The system cannot find the file specified.
```

mgorny:
Okay, adding `%SP_DIR%\` should help.

h-vetinari:
Looking at the list of packages in `%SP_DIR%\torch` (after `pip install`, before we slice it into `libtorch` and `pytorch`)
```
 Directory of C:\bld\libtorch_1737513263999\_h_env\Lib\site-packages\torch\lib

01/22/2025  05:42 AM    <DIR>          .
01/22/2025  05:42 AM    <DIR>          ..
01/22/2025  05:42 AM           358,912 asmjit.dll
01/22/2025  05:42 AM           125,788 asmjit.lib
01/22/2025  05:42 AM           872,448 c10.dll
01/22/2025  05:42 AM           753,338 c10.lib
01/22/2025  05:42 AM           116,214 cpuinfo.lib
01/22/2025  05:42 AM       653,521,082 dnnl.lib
01/22/2025  05:42 AM         4,961,280 fbgemm.dll
01/22/2025  05:42 AM         1,409,606 fbgemm.lib
01/22/2025  05:42 AM         1,497,852 fmt.lib
01/22/2025  05:42 AM        21,732,046 kineto.lib
01/22/2025  05:42 AM           189,296 pthreadpool.lib
01/22/2025  05:42 AM            14,848 shm.dll
01/22/2025  05:42 AM             3,790 shm.lib
01/22/2025  05:42 AM             9,728 torch.dll
01/22/2025  05:42 AM             1,832 torch.lib
01/22/2025  05:42 AM       141,979,136 torch_cpu.dll
01/22/2025  05:42 AM        28,305,954 torch_cpu.lib
01/22/2025  05:42 AM             9,728 torch_global_deps.dll
01/22/2025  05:42 AM        15,196,160 torch_python.dll
01/22/2025  05:42 AM           219,498 torch_python.lib
01/22/2025  05:42 AM         3,698,894 XNNPACK.lib
01/22/2025  05:42 AM             1,924 _C.lib
              22 File(s)    874,979,354 bytes
               2 Dir(s)  68,709,519,360 bytes free
```
I'm pretty sure we also want to move `_C.lib` into site-packages, as the name is definitely not unique or self-explanatory enough for `%LIBRARY_LIB%`. I was mildly surprised that `cpuinfo.lib` and `pthreadpool.lib` don't exist in any other conda-forge packages already, so I guess that's OK.

However, `fmt.lib` looks like it's gonna be a problem, as that directly clashes with `fmt` from https://github.com/conda-forge/fmt-feedstock, which has a pretty finicky ABI too. The format library doesn't appear in the unix outputs of this feedstock, so this is somehow windows-specific.

Not packaging it in either `libtorch` or `pytorch` would work from a packaging standpoint, but not sure that would still work (especially for the JIT stuff). CC @rgommers @isuruf

h-vetinari:
Funnily enough, the windows build already picks up `fmt` from somewhere anyway, even though it's not in the host environment (but it is in the root environment where conda-build gets installed).
```
  -- Found MKL-DNN: TRUE
  -- {fmt} version: 11.0.2
  -- Build type: Release
```

In the linux builds, `fmt` doesn't show up during CMake configuration, but it still gets built:
```
[7464/7474] Linking CXX static library lib/libfmt.a
```
However, that file never ends up getting packaged into any of the outputs, despite
https://github.com/conda-forge/pytorch-cpu-feedstock/blob/4ff92b330141d5d496afdc2d83ca0f32f6adf03d/recipe/build.sh#L206
No idea what's causing that.

> I was mildly surprised that `cpuinfo.lib` and `pthreadpool.lib` don't exist in any other conda-forge packages already, so I guess that's OK.

FWIW, CMake has this in the logs
```
Private Dependencies : Threads::Threads;pthreadpool;cpuinfo;XNNPACK;fbgemm;fp16;gloo;fmt::fmt-header-only;kineto
```
Not sure if the fact that they're considered private would be a strong enough reason to push them to site-packages; I guess we'd then have to always add `%SP_DIR%\torch\lib` to the linker path?

In any case, the `fmt::fmt-header-only` is probably a good indication that we don't actually need `fmt.lib` (but we might need `fmt` as a run-dep for the headers? 🤔).

mgorny:
Interesting enough, CMake and pkg-config files for `fmt` are installed as well. Guessing by:

```
grep: dnnl.lib: binary file matches
```

it's pulled in by `dnnl`. And `dnnl` is mentioned in `torch_cpu.lib` — but I don't know if it's really necessary.

Perhaps we could try removing all `.lib` but `torch`, `torch_cpu` and others we know could be linked directly, and see what happens?

h-vetinari:
> Interesting enough, CMake and pkg-config files for `fmt` are installed as well.

Yeah, we should delete those as well IMO.

> Perhaps we could try removing all `.lib` but `torch`, `torch_cpu` and others we know could be linked directly, and see what happens?

Really not sure, but I guess we can give it a try.

By the way, I found another issue that we should fix here (doesn't just affect windows, bizarrely): #329

h-vetinari:
Looking at the [content](https://conda-metadata-app.streamlit.app/?q=conda-forge%2Fwin-64%2Flibtorch-2.5.1-cuda126_mkl_h0dd7bf4_309.conda&with_broken=true) of the libtorch builds from #316, I also see 
```
Lib/site-packages/torch/lib/cmake/mimalloc-1.8/mimalloc-config-version.cmake
Lib/site-packages/torch/lib/cmake/mimalloc-1.8/mimalloc-config.cmake
Lib/site-packages/torch/lib/cmake/mimalloc-1.8/mimalloc-release.cmake
Lib/site-packages/torch/lib/cmake/mimalloc-1.8/mimalloc.cmake
Lib/site-packages/torch/lib/mimalloc-1.8/mimalloc-static.lib
Lib/site-packages/torch/lib/pkgconfig/mimalloc.pc
```

Not sure why that doesn't show up in the list above. The only reason I can think of is that the `dir %SP_DIR%\torch` was after the `pip install` (i.e. `mimalloc` doesn't get installed there...?), whereas the contents in `libtorch` correspond to the unpacked wheel (which apparently _does_ contain this).

h-vetinari:
> Perhaps we could try removing all `.lib` but `torch`, `torch_cpu` and others we know could be linked directly, and see what happens?

Which libraries do you propose to delete? Out of the [list](https://github.com/conda-forge/pytorch-cpu-feedstock/pull/328#issuecomment-2606409021) above, I can only pick `fmt.lib` as a good candidate.

`dnnl` seems to be used quite heavily in pytorch, even in some of the headers we're providing ([example](https://github.com/pytorch/pytorch/blob/v2.5.1/aten/src/ATen/native/mkldnn/xpu/detail/Utils.h)), so I assume those must be there for anyone compiling against `libtorch` (and same for all the `Private Dependencies` I listed).

isuruf:
`dnnl` would conflict with `onednn` conda package right?

h-vetinari:
> `dnnl` would conflict with `onednn` conda package right?

[Indeed](https://conda-metadata-app.streamlit.app/?q=conda-forge%2Fwin-64%2Fonednn-3.6.2-threadpool_h542669f_0.conda) (I wasn't aware of the contents of this package). That pulls #289 back into focus.

But if a short-term fix is feasible (given that we already published builds and need to fix them), what would you prefer? Move only
```
01/22/2025  05:42 AM             9,728 torch.dll
01/22/2025  05:42 AM             1,832 torch.lib
01/22/2025  05:42 AM       141,979,136 torch_cpu.dll
01/22/2025  05:42 AM        28,305,954 torch_cpu.lib
01/22/2025  05:42 AM             9,728 torch_global_deps.dll
01/22/2025  05:42 AM        15,196,160 torch_python.dll
01/22/2025  05:42 AM           219,498 torch_python.lib
```
to `%LIBRARY_PREFIX%`, leaving the rest in `%SP_DIR%\torch` (and adding some activation to add `-L%SP_DIR%\torch\lib` to `LDFLAGS`)?

h-vetinari:
> and adding some activation to add `-L%SP_DIR%\torch\lib` to `LDFLAGS`

It would probably be better to do this in a way that doesn't leak to the outside, à la
`target_link_libraries(... INTERFACE %SP_DIR%\lib\...)`

mgorny:
Or perhaps we could keep everything in `%SP_DIR%`, and install some CMake wrapper into `%LIBRARY_PREFIX%` that would point other packages at the correct location?

Hmm, also I've also noticed that `asmjit.dll` and `fbgemm.dll` are both in `bin\` and `lib\` subdirectories, but other DLLs are only in `lib\`.

mgorny:
I'm talking of https://conda-metadata-app.streamlit.app/?q=conda-forge%2Fwin-64%2Flibtorch-2.5.1-cpu_mkl_hbbd3bdd_109.conda#files. Note that there are both:

```
Lib/site-packages/torch/bin/asmjit.dll
Lib/site-packages/torch/bin/fbgemm.dll
[…]
Lib/site-packages/torch/lib/asmjit.dll
[…]
Lib/site-packages/torch/lib/fbgemm.dll
```

h-vetinari:
Alright, at least we're back to passing the import tests, now only failing two tests:
```
=========================== short test summary info ===========================
FAILED [0.2329s] test/test_autograd.py::TestAutograd::test_multi_grad_all_hooks - RuntimeError: Error building extension 'test_autograd_cpp_node'

    python test\test_autograd.py TestAutograd.test_multi_grad_all_hooks

FAILED [0.1248s] test/test_custom_ops.py::TestCustomOp::test_autograd_function_backed_op - RuntimeError: Error building extension 'mylib'

    python test\test_custom_ops.py TestCustomOp.test_autograd_function_backed_op

= 2 failed, 7484 passed, 1432 skipped, 43 deselected, 31 xfailed, 75940 warnings in 1017.41s (0:16:57) =
```
both related to building custom extensions, so presumably broken by us moving stuff around.

<details>

```
================================== FAILURES ===================================
___________________ TestAutograd.test_multi_grad_all_hooks ____________________

build_directory = 'C:\\Users\\runnerx.cirun-7699dd3\\AppData\\Local\\torch_extensions\\torch_extensions\\Cache\\py311_cpu\\test_autograd_cpp_node'
verbose = True
error_prefix = "Error building extension 'test_autograd_cpp_node'"

    def _run_ninja_build(build_directory: str, verbose: bool, error_prefix: str) -> None:
        command = ['ninja', '-v']
        num_workers = _get_num_workers(verbose)
        if num_workers is not None:
            command.extend(['-j', str(num_workers)])
        env = os.environ.copy()
        # Try to activate the vc env for the users
        if IS_WINDOWS and 'VSCMD_ARG_TGT_ARCH' not in env:
            from setuptools import distutils
    
            plat_name = distutils.util.get_platform()
            plat_spec = PLAT_TO_VCVARS[plat_name]
    
            vc_env = distutils._msvccompiler._get_vc_env(plat_spec)
            vc_env = {k.upper(): v for k, v in vc_env.items()}
            for k, v in env.items():
                uk = k.upper()
                if uk not in vc_env:
                    vc_env[uk] = v
            env = vc_env
        try:
            sys.stdout.flush()
            sys.stderr.flush()
            # Warning: don't pass stdout=None to subprocess.run to get output.
            # subprocess.run assumes that sys.__stdout__ has not been modified and
            # attempts to write to it by default.  However, when we call _run_ninja_build
            # from ahead-of-time cpp extensions, the following happens:
            # 1) If the stdout encoding is not utf-8, setuptools detachs __stdout__.
            #    https://github.com/pypa/setuptools/blob/7e97def47723303fafabe48b22168bbc11bb4821/setuptools/dist.py#L1110
            #    (it probably shouldn't do this)
            # 2) subprocess.run (on POSIX, with no stdout override) relies on
            #    __stdout__ not being detached:
            #    https://github.com/python/cpython/blob/c352e6c7446c894b13643f538db312092b351789/Lib/subprocess.py#L1214
            # To work around this, we pass in the fileno directly and hope that
            # it is valid.
            stdout_fileno = 1
>           subprocess.run(
                command,
                stdout=stdout_fileno if verbose else subprocess.PIPE,
                stderr=subprocess.STDOUT,
                cwd=build_directory,
                check=True,
                env=env)

..\_test_env\Lib\site-packages\torch\utils\cpp_extension.py:2104: 

input = None, capture_output = False, timeout = None, check = True
popenargs = (['ninja', '-v'],)
kwargs = {'cwd': 'C:\\Users\\runnerx.cirun-7699dd3\\AppData\\Local\\torch_extensions\\torch_extensions\\Cache\\py311_cpu\\test_...ws', 'ALLUSERSPROFILE': 'C:\\ProgramData', 'ANDROID_HOME': 'C:\\Android\\android-sdk', ...}, 'stderr': -2, 'stdout': 1}
process = <Popen: returncode: 1 args: ['ninja', '-v']>, stdout = None
stderr = None, retcode = 1

    def run(*popenargs,
            input=None, capture_output=False, timeout=None, check=False, **kwargs):
        """Run command with arguments and return a CompletedProcess instance.
    
        The returned instance will have attributes args, returncode, stdout and
        stderr. By default, stdout and stderr are not captured, and those attributes
        will be None. Pass stdout=PIPE and/or stderr=PIPE in order to capture them,
        or pass capture_output=True to capture both.
    
        If check is True and the exit code was non-zero, it raises a
        CalledProcessError. The CalledProcessError object will have the return code
        in the returncode attribute, and output & stderr attributes if those streams
        were captured.
    
        If timeout is given, and the process takes too long, a TimeoutExpired
        exception will be raised.
    
        There is an optional argument "input", allowing you to
        pass bytes or a string to the subprocess's stdin.  If you use this argument
        you may not also use the Popen constructor's "stdin" argument, as
        it will be used internally.
    
        By default, all communication is in bytes, and therefore any "input" should
        be bytes, and the stdout and stderr will be bytes. If in text mode, any
        "input" should be a string, and stdout and stderr will be strings decoded
        according to locale encoding, or by "encoding" if set. Text mode is
        triggered by setting any of text, encoding, errors or universal_newlines.
    
        The other arguments are the same as for the Popen constructor.
        """
        if input is not None:
            if kwargs.get('stdin') is not None:
                raise ValueError('stdin and input arguments may not both be used.')
            kwargs['stdin'] = PIPE
    
        if capture_output:
            if kwargs.get('stdout') is not None or kwargs.get('stderr') is not None:
                raise ValueError('stdout and stderr arguments may not be used '
                                 'with capture_output.')
            kwargs['stdout'] = PIPE
            kwargs['stderr'] = PIPE
    
        with Popen(*popenargs, **kwargs) as process:
            try:
                stdout, stderr = process.communicate(input, timeout=timeout)
            except TimeoutExpired as exc:
                process.kill()
                if _mswindows:
                    # Windows accumulates the output in a single blocking
                    # read() call run on child threads, with the timeout
                    # being done in a join() on those threads.  communicate()
                    # _after_ kill() is required to collect that and add it
                    # to the exception.
                    exc.stdout, exc.stderr = process.communicate()
                else:
                    # POSIX _communicate already populated the output so
                    # far into the TimeoutExpired exception.
                    process.wait()
                raise
            except:  # Including KeyboardInterrupt, communicate handled that.
                process.kill()
                # We don't call process.wait() as .__exit__ does that for us.
                raise
            retcode = process.poll()
            if check and retcode:
>               raise CalledProcessError(retcode, process.args,
                                         output=stdout, stderr=stderr)
E               subprocess.CalledProcessError: Command '['ninja', '-v']' returned non-zero exit status 1.

..\_test_env\Lib\subprocess.py:571: CalledProcessError

The above exception was the direct cause of the following exception:

self = <test_autograd.TestAutograd testMethod=test_multi_grad_all_hooks>

        def test_multi_grad_all_hooks(self):
            t1 = torch.rand(2, requires_grad=True)
            t2 = torch.rand(2, requires_grad=True)
            t3 = torch.rand(2, requires_grad=True)
            t4 = torch.rand(2, requires_grad=True)
    
            # Ensure we properly detect all types of Nodes here
            # C++ Node
            t1 = t1.mul(2)
    
            # Python custom Function
            class Foo(Function):
                @staticmethod
                def forward(ctx, a):
                    return a.clone()
    
                @staticmethod
                def backward(ctx, gO):
                    return gO
    
            t2 = Foo.apply(t2)
    
            # C++ Node
            t3 = torch._C._functions.UndefinedGrad()(t3)
    
            # C++ Custom Op
            cpp_source = """
    struct CustomOpAutogradFunction : public torch::autograd::Function<CustomOpAutogradFunction> {
      static torch::Tensor forward(
          torch::autograd::AutogradContext* ctx,
          const torch::Tensor& x) {
        return x.clone();
      }
    
      static torch::autograd::variable_list backward(
          torch::autograd::AutogradContext *ctx,
          torch::autograd::variable_list grad_output) {
        return grad_output;
      }
    };
    
    torch::Tensor custom_op_backed_by_autograd_fn(torch::Tensor x) {
      return CustomOpAutogradFunction::apply(x);
    }
    
    TORCH_LIBRARY(test_autograd_cpp_node, m) {
        m.def("custom_op_backed_by_autograd_fn", custom_op_backed_by_autograd_fn);
    }
            """
    
>           module = load_inline(
                name="test_autograd_cpp_node",
                cpp_sources=cpp_source,
                functions="custom_op_backed_by_autograd_fn",
                verbose=True,
            )

test\test_autograd.py:9864: 
..\_test_env\Lib\site-packages\torch\utils\cpp_extension.py:1646: in load_inline
    return _jit_compile(
..\_test_env\Lib\site-packages\torch\utils\cpp_extension.py:1721: in _jit_compile
    _write_ninja_file_and_build_library(
..\_test_env\Lib\site-packages\torch\utils\cpp_extension.py:1833: in _write_ninja_file_and_build_library
    _run_ninja_build(

build_directory = 'C:\\Users\\runnerx.cirun-7699dd3\\AppData\\Local\\torch_extensions\\torch_extensions\\Cache\\py311_cpu\\test_autograd_cpp_node'
verbose = True
error_prefix = "Error building extension 'test_autograd_cpp_node'"

    def _run_ninja_build(build_directory: str, verbose: bool, error_prefix: str) -> None:
        command = ['ninja', '-v']
        num_workers = _get_num_workers(verbose)
        if num_workers is not None:
            command.extend(['-j', str(num_workers)])
        env = os.environ.copy()
        # Try to activate the vc env for the users
        if IS_WINDOWS and 'VSCMD_ARG_TGT_ARCH' not in env:
            from setuptools import distutils
    
            plat_name = distutils.util.get_platform()
            plat_spec = PLAT_TO_VCVARS[plat_name]
    
            vc_env = distutils._msvccompiler._get_vc_env(plat_spec)
            vc_env = {k.upper(): v for k, v in vc_env.items()}
            for k, v in env.items():
                uk = k.upper()
                if uk not in vc_env:
                    vc_env[uk] = v
            env = vc_env
        try:
            sys.stdout.flush()
            sys.stderr.flush()
            # Warning: don't pass stdout=None to subprocess.run to get output.
            # subprocess.run assumes that sys.__stdout__ has not been modified and
            # attempts to write to it by default.  However, when we call _run_ninja_build
            # from ahead-of-time cpp extensions, the following happens:
            # 1) If the stdout encoding is not utf-8, setuptools detachs __stdout__.
            #    https://github.com/pypa/setuptools/blob/7e97def47723303fafabe48b22168bbc11bb4821/setuptools/dist.py#L1110
            #    (it probably shouldn't do this)
            # 2) subprocess.run (on POSIX, with no stdout override) relies on
            #    __stdout__ not being detached:
            #    https://github.com/python/cpython/blob/c352e6c7446c894b13643f538db312092b351789/Lib/subprocess.py#L1214
            # To work around this, we pass in the fileno directly and hope that
            # it is valid.
            stdout_fileno = 1
            subprocess.run(
                command,
                stdout=stdout_fileno if verbose else subprocess.PIPE,
                stderr=subprocess.STDOUT,
                cwd=build_directory,
                check=True,
                env=env)
        except subprocess.CalledProcessError as e:
            # Python 2 and 3 compatible way of getting the error object.
            _, error, _ = sys.exc_info()
            # error.output contains the stdout and stderr of the build attempt.
            message = error_prefix
            # `error` is a CalledProcessError (which has an `output`) attribute, but
            # mypy thinks it's Optional[BaseException] and doesn't narrow
            if hasattr(error, 'output') and error.output:  # type: ignore[union-attr]
                message += f": {error.output.decode(*SUBPROCESS_DECODE_ARGS)}"  # type: ignore[union-attr]
>           raise RuntimeError(message) from e
E           RuntimeError: Error building extension 'test_autograd_cpp_node'
E           
E           To execute this test, run the following from the base repo dir:
E               python test\test_autograd.py TestAutograd.test_multi_grad_all_hooks
E           
E           This message can be suppressed by setting PYTORCH_PRINT_REPRO_ON_FAILURE=0

..\_test_env\Lib\site-packages\torch\utils\cpp_extension.py:2120: RuntimeError
________________ TestCustomOp.test_autograd_function_backed_op ________________

build_directory = 'C:\\Users\\runnerx.cirun-7699dd3\\AppData\\Local\\torch_extensions\\torch_extensions\\Cache\\py311_cpu\\mylib'
verbose = True, error_prefix = "Error building extension 'mylib'"

    def _run_ninja_build(build_directory: str, verbose: bool, error_prefix: str) -> None:
        command = ['ninja', '-v']
        num_workers = _get_num_workers(verbose)
        if num_workers is not None:
            command.extend(['-j', str(num_workers)])
        env = os.environ.copy()
        # Try to activate the vc env for the users
        if IS_WINDOWS and 'VSCMD_ARG_TGT_ARCH' not in env:
            from setuptools import distutils
    
            plat_name = distutils.util.get_platform()
            plat_spec = PLAT_TO_VCVARS[plat_name]
    
            vc_env = distutils._msvccompiler._get_vc_env(plat_spec)
            vc_env = {k.upper(): v for k, v in vc_env.items()}
            for k, v in env.items():
                uk = k.upper()
                if uk not in vc_env:
                    vc_env[uk] = v
            env = vc_env
        try:
            sys.stdout.flush()
            sys.stderr.flush()
            # Warning: don't pass stdout=None to subprocess.run to get output.
            # subprocess.run assumes that sys.__stdout__ has not been modified and
            # attempts to write to it by default.  However, when we call _run_ninja_build
            # from ahead-of-time cpp extensions, the following happens:
            # 1) If the stdout encoding is not utf-8, setuptools detachs __stdout__.
            #    https://github.com/pypa/setuptools/blob/7e97def47723303fafabe48b22168bbc11bb4821/setuptools/dist.py#L1110
            #    (it probably shouldn't do this)
            # 2) subprocess.run (on POSIX, with no stdout override) relies on
            #    __stdout__ not being detached:
            #    https://github.com/python/cpython/blob/c352e6c7446c894b13643f538db312092b351789/Lib/subprocess.py#L1214
            # To work around this, we pass in the fileno directly and hope that
            # it is valid.
            stdout_fileno = 1
>           subprocess.run(
                command,
                stdout=stdout_fileno if verbose else subprocess.PIPE,
                stderr=subprocess.STDOUT,
                cwd=build_directory,
                check=True,
                env=env)

..\_test_env\Lib\site-packages\torch\utils\cpp_extension.py:2104: 

input = None, capture_output = False, timeout = None, check = True
popenargs = (['ninja', '-v'],)
kwargs = {'cwd': 'C:\\Users\\runnerx.cirun-7699dd3\\AppData\\Local\\torch_extensions\\torch_extensions\\Cache\\py311_cpu\\mylib...ws', 'ALLUSERSPROFILE': 'C:\\ProgramData', 'ANDROID_HOME': 'C:\\Android\\android-sdk', ...}, 'stderr': -2, 'stdout': 1}
process = <Popen: returncode: 1 args: ['ninja', '-v']>, stdout = None
stderr = None, retcode = 1

    def run(*popenargs,
            input=None, capture_output=False, timeout=None, check=False, **kwargs):
        """Run command with arguments and return a CompletedProcess instance.
    
        The returned instance will have attributes args, returncode, stdout and
        stderr. By default, stdout and stderr are not captured, and those attributes
        will be None. Pass stdout=PIPE and/or stderr=PIPE in order to capture them,
        or pass capture_output=True to capture both.
    
        If check is True and the exit code was non-zero, it raises a
        CalledProcessError. The CalledProcessError object will have the return code
        in the returncode attribute, and output & stderr attributes if those streams
        were captured.
    
        If timeout is given, and the process takes too long, a TimeoutExpired
        exception will be raised.
    
        There is an optional argument "input", allowing you to
        pass bytes or a string to the subprocess's stdin.  If you use this argument
        you may not also use the Popen constructor's "stdin" argument, as
        it will be used internally.
    
        By default, all communication is in bytes, and therefore any "input" should
        be bytes, and the stdout and stderr will be bytes. If in text mode, any
        "input" should be a string, and stdout and stderr will be strings decoded
        according to locale encoding, or by "encoding" if set. Text mode is
        triggered by setting any of text, encoding, errors or universal_newlines.
    
        The other arguments are the same as for the Popen constructor.
        """
        if input is not None:
            if kwargs.get('stdin') is not None:
                raise ValueError('stdin and input arguments may not both be used.')
            kwargs['stdin'] = PIPE
    
        if capture_output:
            if kwargs.get('stdout') is not None or kwargs.get('stderr') is not None:
                raise ValueError('stdout and stderr arguments may not be used '
                                 'with capture_output.')
            kwargs['stdout'] = PIPE
            kwargs['stderr'] = PIPE
    
        with Popen(*popenargs, **kwargs) as process:
            try:
                stdout, stderr = process.communicate(input, timeout=timeout)
            except TimeoutExpired as exc:
                process.kill()
                if _mswindows:
                    # Windows accumulates the output in a single blocking
                    # read() call run on child threads, with the timeout
                    # being done in a join() on those threads.  communicate()
                    # _after_ kill() is required to collect that and add it
                    # to the exception.
                    exc.stdout, exc.stderr = process.communicate()
                else:
                    # POSIX _communicate already populated the output so
                    # far into the TimeoutExpired exception.
                    process.wait()
                raise
            except:  # Including KeyboardInterrupt, communicate handled that.
                process.kill()
                # We don't call process.wait() as .__exit__ does that for us.
                raise
            retcode = process.poll()
            if check and retcode:
>               raise CalledProcessError(retcode, process.args,
                                         output=stdout, stderr=stderr)
E               subprocess.CalledProcessError: Command '['ninja', '-v']' returned non-zero exit status 1.

..\_test_env\Lib\subprocess.py:571: CalledProcessError

The above exception was the direct cause of the following exception:

self = <test_custom_ops.TestCustomOp testMethod=test_autograd_function_backed_op>

        def test_autograd_function_backed_op(self):
            cpp_source = """
    struct CustomOpAutogradFunction : public torch::autograd::Function<CustomOpAutogradFunction> {
      static constexpr bool is_traceable = true;
    
      static torch::Tensor forward(
          torch::autograd::AutogradContext* ctx,
          const torch::Tensor& x) {
        return x;
      }
    
      static torch::autograd::variable_list backward(
          torch::autograd::AutogradContext *ctx,
          torch::autograd::variable_list grad_output) {
        return grad_output;
      }
    };
    
    torch::Tensor custom_op_backed_by_autograd_fn(const torch::Tensor& x) {
      return CustomOpAutogradFunction::apply(x);
    }
    
    TORCH_LIBRARY(mylib, m) {
        m.def("custom_op_backed_by_autograd_fn", custom_op_backed_by_autograd_fn);
    }
            """
    
>           module = torch.utils.cpp_extension.load_inline(
                name="mylib",
                cpp_sources=cpp_source,
                functions="custom_op_backed_by_autograd_fn",
                verbose=True,
            )

test\test_custom_ops.py:2080: 
..\_test_env\Lib\site-packages\torch\utils\cpp_extension.py:1646: in load_inline
    return _jit_compile(
..\_test_env\Lib\site-packages\torch\utils\cpp_extension.py:1721: in _jit_compile
    _write_ninja_file_and_build_library(
..\_test_env\Lib\site-packages\torch\utils\cpp_extension.py:1833: in _write_ninja_file_and_build_library
    _run_ninja_build(

build_directory = 'C:\\Users\\runnerx.cirun-7699dd3\\AppData\\Local\\torch_extensions\\torch_extensions\\Cache\\py311_cpu\\mylib'
verbose = True, error_prefix = "Error building extension 'mylib'"

    def _run_ninja_build(build_directory: str, verbose: bool, error_prefix: str) -> None:
        command = ['ninja', '-v']
        num_workers = _get_num_workers(verbose)
        if num_workers is not None:
            command.extend(['-j', str(num_workers)])
        env = os.environ.copy()
        # Try to activate the vc env for the users
        if IS_WINDOWS and 'VSCMD_ARG_TGT_ARCH' not in env:
            from setuptools import distutils
    
            plat_name = distutils.util.get_platform()
            plat_spec = PLAT_TO_VCVARS[plat_name]
    
            vc_env = distutils._msvccompiler._get_vc_env(plat_spec)
            vc_env = {k.upper(): v for k, v in vc_env.items()}
            for k, v in env.items():
                uk = k.upper()
                if uk not in vc_env:
                    vc_env[uk] = v
            env = vc_env
        try:
            sys.stdout.flush()
            sys.stderr.flush()
            # Warning: don't pass stdout=None to subprocess.run to get output.
            # subprocess.run assumes that sys.__stdout__ has not been modified and
            # attempts to write to it by default.  However, when we call _run_ninja_build
            # from ahead-of-time cpp extensions, the following happens:
            # 1) If the stdout encoding is not utf-8, setuptools detachs __stdout__.
            #    https://github.com/pypa/setuptools/blob/7e97def47723303fafabe48b22168bbc11bb4821/setuptools/dist.py#L1110
            #    (it probably shouldn't do this)
            # 2) subprocess.run (on POSIX, with no stdout override) relies on
            #    __stdout__ not being detached:
            #    https://github.com/python/cpython/blob/c352e6c7446c894b13643f538db312092b351789/Lib/subprocess.py#L1214
            # To work around this, we pass in the fileno directly and hope that
            # it is valid.
            stdout_fileno = 1
            subprocess.run(
                command,
                stdout=stdout_fileno if verbose else subprocess.PIPE,
                stderr=subprocess.STDOUT,
                cwd=build_directory,
                check=True,
                env=env)
        except subprocess.CalledProcessError as e:
            # Python 2 and 3 compatible way of getting the error object.
            _, error, _ = sys.exc_info()
            # error.output contains the stdout and stderr of the build attempt.
            message = error_prefix
            # `error` is a CalledProcessError (which has an `output`) attribute, but
            # mypy thinks it's Optional[BaseException] and doesn't narrow
            if hasattr(error, 'output') and error.output:  # type: ignore[union-attr]
                message += f": {error.output.decode(*SUBPROCESS_DECODE_ARGS)}"  # type: ignore[union-attr]
>           raise RuntimeError(message) from e
E           RuntimeError: Error building extension 'mylib'
E           
E           To execute this test, run the following from the base repo dir:
E               python test\test_custom_ops.py TestCustomOp.test_autograd_function_backed_op
E           
E           This message can be suppressed by setting PYTORCH_PRINT_REPRO_ON_FAILURE=0

..\_test_env\Lib\site-packages\torch\utils\cpp_extension.py:2120: RuntimeError
============================== warnings summary ===============================
```

</details>

h-vetinari:
> I'm talking of https://conda-metadata-app.streamlit.app/?q=conda-forge%2Fwin-64%2Flibtorch-2.5.1-cpu_mkl_hbbd3bdd_109.conda#files.

OK, gotcha, but that won't be relevant anymore as of this PR (because we've revamped that part substantially, and now we're splitting `.lib` and `.dll` more consequently)

mgorny:
Tests failed due to mismatched file list, so updated that now. I've also added a fix for #329.

h-vetinari:
> I've also added a fix for #329.

https://github.com/conda-forge/pytorch-cpu-feedstock/pull/328/commits/fc5693c26e749f09f5cf314c4d77b32ce2290248 was the fix for that. Your commit tackles the `libtorch` build and is superfluous AFAICT. Can you please remove that commit for now?

recipe/bld.bat
```diff
@@ -175,46 +172,61 @@ if errorlevel 1 exit /b 1
...
     @REM Keep the original backed up to sed later
     copy build\CMakeCache.txt build\CMakeCache.txt.orig
+    if %ERRORLEVEL% neq 0 exit 1
 ) else if "%PKG_NAME%" == "pytorch" (
+    @REM Move libtorch_python and remove the other directories afterwards.
+    robocopy /NP /NFL /NDL /NJH /E %SP_DIR%\torch\lib\ %LIBRARY_BIN%\ torch_python.dll
+    robocopy /NP /NFL /NDL /NJH /E %SP_DIR%\torch\lib\ %LIBRARY_LIB%\ torch_python.lib
+    robocopy /NP /NFL /NDL /NJH /E %SP_DIR%\torch\lib\ %LIBRARY_LIB%\ _C.lib
+    rmdir /s /q %SP_DIR%\torch\lib
     rmdir /s /q %SP_DIR%\torch\bin
```
h-vetinari, line 220: Just realized that we need to modify this bit now to actually install the remaining libraries. I need to jump now, will do it in ~2-3h otherwise.

h-vetinari:
Well, that last commit to see the internals of the failed call to `ninja` was a wash. Not sure if this is just missing a flush, but in any case, it's not obvious where this is going wrong. I looked at the sources of [the two](https://github.com/pytorch/pytorch/blob/v2.5.1/test/test_autograd.py#L9840-L9862) [failing](https://github.com/pytorch/pytorch/blob/v2.5.1/test/test_custom_ops.py#L2054-L2078) tests, and it's not obvious what issues they could encounter - certainly it doesn't look like they're using exotic symbols, but I guess we can always run into missing transitive dependencies that require some of the static libs we've removed (if that is even the problem).

h-vetinari:
OK, so I downloaded the previous pytorch cpu build, set up the test environment, and was able to run both tests successfully
```
(test) E:\upstreams\pytorch>python test\test_autograd.py TestAutograd.test_multi_grad_all_hooks
Using C:\Users\[...]\AppData\Local\torch_extensions\torch_extensions\Cache\py312_cpu as PyTorch extensions root...
Emitting ninja build file C:\Users\[...]\AppData\Local\torch_extensions\torch_extensions\Cache\py312_cpu\test_autograd_cpp_node\build.ninja...
Building extension module test_autograd_cpp_node...
Allowing ninja to set a default number of workers... (overridable by setting the environment variable MAX_JOBS=N)
[1/2] cl /showIncludes -DTORCH_EXTENSION_NAME=test_autograd_cpp_node -DTORCH_API_INCLUDE_EXTENSION_H -IE:\miniforge\envs\test\Lib\site-packages\torch\include -IE:\miniforge\envs\test\Lib\site-packages\torch\include\torch\csrc\api\include -IE:\miniforge\envs\test\Lib\site-packages\torch\include\TH -IE:\miniforge\envs\test\Lib\site-packages\torch\include\THC -IE:\miniforge\envs\test\Include -D_GLIBCXX_USE_CXX11_ABI=0 /MD /wd4819 /wd4251 /wd4244 /wd4267 /wd4275 /wd4018 /wd4190 /wd4624 /wd4067 /wd4068 /EHsc /std:c++17 -c C:\Users\[...]\AppData\Local\torch_extensions\torch_extensions\Cache\py312_cpu\test_autograd_cpp_node\main.cpp /Fomain.o
Microsoft (R) C/C++ Optimizing Compiler Version 19.29.30158 for x64
Copyright (C) Microsoft Corporation.  All rights reserved.

[2/2] "C:\Program Files\Microsoft Visual Studio\2022\Community\VC\Tools\MSVC\14.29.30133\bin\HostX64\x64/link.exe" main.o /nologo /DLL c10.lib torch_cpu.lib torch.lib /LIBPATH:E:\miniforge\envs\test\Lib\site-packages\torch\lib torch_python.lib /LIBPATH:E:\miniforge\envs\test\libs /out:test_autograd_cpp_node.pyd
   Creating library test_autograd_cpp_node.lib and object test_autograd_cpp_node.exp
Loading extension module test_autograd_cpp_node...
.
----------------------------------------------------------------------
Ran 1 test in 10.331s

OK

(test) E:\upstreams\pytorch>python test\test_custom_ops.py TestCustomOp.test_autograd_function_backed_op
Using C:\Users\[...]\AppData\Local\torch_extensions\torch_extensions\Cache\py312_cpu as PyTorch extensions root...
Creating extension directory C:\Users\[...]\AppData\Local\torch_extensions\torch_extensions\Cache\py312_cpu\mylib...
Emitting ninja build file C:\Users\[...]\AppData\Local\torch_extensions\torch_extensions\Cache\py312_cpu\mylib\build.ninja...
Building extension module mylib...
Allowing ninja to set a default number of workers... (overridable by setting the environment variable MAX_JOBS=N)
[1/2] cl /showIncludes -DTORCH_EXTENSION_NAME=mylib -DTORCH_API_INCLUDE_EXTENSION_H -IE:\miniforge\envs\test\Lib\site-packages\torch\include -IE:\miniforge\envs\test\Lib\site-packages\torch\include\torch\csrc\api\include -IE:\miniforge\envs\test\Lib\site-packages\torch\include\TH -IE:\miniforge\envs\test\Lib\site-packages\torch\include\THC -IE:\miniforge\envs\test\Include -D_GLIBCXX_USE_CXX11_ABI=0 /MD /wd4819 /wd4251 /wd4244 /wd4267 /wd4275 /wd4018 /wd4190 /wd4624 /wd4067 /wd4068 /EHsc /std:c++17 -c C:\Users\[...]\AppData\Local\torch_extensions\torch_extensions\Cache\py312_cpu\mylib\main.cpp /Fomain.o
Microsoft (R) C/C++ Optimizing Compiler Version 19.29.30158 for x64
Copyright (C) Microsoft Corporation.  All rights reserved.

[2/2] "C:\Program Files\Microsoft Visual Studio\2022\Community\VC\Tools\MSVC\14.29.30133\bin\HostX64\x64/link.exe" main.o /nologo /DLL c10.lib torch_cpu.lib torch.lib /LIBPATH:E:\miniforge\envs\test\Lib\site-packages\torch\lib torch_python.lib /LIBPATH:E:\miniforge\envs\test\libs /out:mylib.pyd
   Creating library mylib.lib and object mylib.exp
Loading extension module mylib...
.
----------------------------------------------------------------------
Ran 1 test in 12.550s

OK
```

Then I moved all the DLLs to `%LIBRARY_PREFIX%\bin`, deleted the `torch_extensions` cache, and ran it again - also ran fine.

Note though the `/LIBPATH:E:\miniforge\envs\test\Lib\site-packages\torch\lib` for the linker - IOW, this is where the test expects the `.lib` files. The reason our windows compilers don't check in `%LIBRARY_LIB%` by default is that this is only [done](https://github.com/conda-forge/vc-feedstock/blob/b2c4a5f5b92936da6df87cc2fe28d7001487cec3/recipe/activate.bat#L63-L67) when building packages (which isn't the case during the test phase). This is in contrast to what our [unix compilers do](https://github.com/conda-forge/ctng-compiler-activation-feedstock/blob/987224ce91e63faf248a07619009a3049034d8e4/recipe/activate-gcc.sh#L89-L104), which also explains why moving everything to `$PREFIX/lib` works there.

For completeness, I also tried again after deleting the static libraries (not the import libraries for the DLLs), and again these two tests ran fine (so deleting `fmt.lib` wasn't the issue as I had feared).

I think the nicest approach here would be to modify our compiler activation on windows to look into `%LIBRARY_LIB%` by default. For that I opened https://github.com/conda-forge/vc-feedstock/pull/90, let's see how that goes. If that is not acceptable for some reason, then the easiest will be to just move the DLLs, and leave the `.lib`s in `%SP_DIR%\torch\lib` (even if that means we might have to patch downstream packages to look there specifically).

h-vetinari:
OK, with https://github.com/conda-forge/vc-feedstock/pull/90, I can now move all the `.lib`s to `%LIBRARY_LIB%` and the tests still work. However, what didn't work was moving the headers to `%LIBRARY_INC%`. That's because the torch headers actually have a nested structure:
```
Lib/site-packages/torch/include/torch/csrc/api/include/torch/all.h
                        ^^^^^^^                ^^^^^^^
                     entry point 1          entry point 2
```
And just moving `torch/csrc/api/include/torch/all.h` to `%LIBRARY_INC%` will not resolve `#include <torch/all.h>`. Pytorch generally adds [both](https://github.com/pytorch/pytorch/blob/v2.5.1/cmake/TorchConfig.cmake.in#L61) paths to the includes, and so [does](https://github.com/pytorch/pytorch/blob/v2.5.1/torch/utils/cpp_extension.py#L1161) the test infrastructure.

I don't understand how this works out on unix TBH (where we're moving the headers to `$PREFIX/include`, but not handling `torch/csrc/api/include` explicitly AFAICT). Maybe it somehow gets persisted in [`INTERFACE_INCLUDE_DIRECTORIES`](https://github.com/pytorch/pytorch/blob/v2.5.1/cmake/TorchConfig.cmake.in#L159)?

Anyway, I think I'll try patching this, unless someone has better ideas?

mgorny:
> I don't understand how this works out on unix TBH (where we're moving the headers to `$PREFIX/include`, but not handling `torch/csrc/api/include` explicitly AFAICT). Maybe it somehow gets persisted in [`INTERFACE_INCLUDE_DIRECTORIES`](https://github.com/pytorch/pytorch/blob/v2.5.1/cmake/TorchConfig.cmake.in#L159)?

On Linux, we still have a bunch of headers in `${SP_DIR}/torch/include` and symlinks for the directories moved to `${CONDA_PREFIX}/include`.

> Anyway, I think I'll try patching this, unless someone has better ideas?

Patching sounds good to me. For CMake, I guess changing `TORCH_INSTALL_PREFIX` should work. However, given that they explicitly repeat it in tests, I'm worried that other Python packages also hardwire these paths.

h-vetinari:
> symlinks for the directories moved to `${CONDA_PREFIX}/include`.

Ah right, forgot the symlink!

> For CMake, I guess changing `TORCH_INSTALL_PREFIX` should work. However, given that they explicitly repeat it in tests, I'm worried that other Python packages also hardwire these paths.

We might need to do both? For now I'd let this run through and see what comes out. Feel free to push more fixes, I'll be offline for the next ~8h.

recipe/patches/0017-point-include-paths-to-PREFIX-include.patch
```diff
@@ -0,0 +1,33 @@
...
++++ b/torch/utils/cpp_extension.py
+@@ -1155,10 +1155,19 @@ def include_paths(cuda: bool = False) -> List[str]:
+         A list of include path strings.
+     """
+     lib_include = os.path.join(_TORCH_PATH, 'include')
++    if os.environ.get("CONDA_BUILD", None) is not None:
++        pieces = [os.environ["PREFIX"]] + IS_WINDOWS * ["Library"] + ["include"]
++        lib_include = os.path.join(*pieces)
++    elif os.environ.get("CONDA_PREFIX", None) is not None:
++        pieces = [os.environ["CONDA_PREFIX"]] + IS_WINDOWS * ["Library"] + ["include"]
```
mgorny, line 22: I wish we wouldn't have to be explicit about that `"Library"`.

mgorny:
I guess it would make sense to remove the symlinks from Linux too, for consistency.

hmaarrfk:
I’m worried about moving too many files around without patching. I’m too busy in the coming weeks to read through the bottom thread but here it is for reference

https://github.com/pytorch/pytorch/issues/144690

I’ll remind us that the main goal of having the so in our shared library is to avoid conda packaging from giving downstream packages the need to think about whether or not they have to add something to the missing dso whitelist.

recipe/patches/0017-point-include-paths-to-PREFIX-include.patch
h-vetinari: Perhaps one day we can get rid of it... https://github.com/conda-forge/conda-forge.github.io/issues/2295

h-vetinari:
Phew, I'm glad this finally passed. #329 still isn't solved, let me reinstate your a514e05f0d0a26d53532357e16a31a1cccfa9dbb.

I'm fine to keep the symlink on linux - it makes life easier, but yeah, the lack of symlinks on windows will probably bite use in a few ways here. I'm mostly keen to get this PR merged and check again what happens when building torchvision etc.

At least for now we've matched the design from the unix side pretty closely (modulo the symlinks) in how the libs are all in `%LIBRARY_{LIB,BIN}%`, and the `ATen, c10, shm, torch` headers are in `%LIBRARY_INc%`, while the rest stays in `%SP_DIR%\torch\include`.

mgorny:
Don't you want to enable the remaining Windows builds? I wanted to add removal of `TH_BINARY_BUILD` too. If that's fine, I'll push it along with rerender to restore all build targets.

h-vetinari:
> Don't you want to enable the remaining Windows builds?

Of course, but I wanted to solve all the current issues first, and also ask @hmaarrfk and @isuruf for review, if they have time. CC also @danpetry if interested.

> I wanted to add removal of `TH_BINARY_BUILD` too. If that's fine, I'll push it along with rerender to restore all build targets.

How high is your confidence that this will work first try? I think this PR is already doing a lot, and I'd like to get it out so we can start building dependent packages and test out whether things work as intended. I wouldn't be surprised if follow-ups turn out to be necessary anyway, so I'd prefer to put `TH_BINARY_BUILD` into the next round.

h-vetinari:
FWIW, #329 is now fixed as well, so I think this is ready (modulo a revert of 3ebf1a9684a23d971b582d88d851948c134e6d50 + a rerender, which I'm holding off on for now, in case we need some more rounds after feedback).

mgorny:
I'd still do the `TH_BINARY_BUILD` thing, but perhaps after all builds pass, so we don't bump the revision twice (or we can defer that to anaconda sync PR) — i.e. basically add it, if CI passes, merge with it, if CI fails, merge the previous commit.

h-vetinari:
Can you raise a PR on your fork against this branch? If you enable actions on your repo, this should actually run on the open-gpu server (at least it does by default on my fork). That way we can keep the CI runs separate and don't have to roll back if it goes wrong.

mgorny:
> Can you raise a PR on your fork against this branch? If you enable actions on your repo, this should actually run on the open-gpu server (at least it does by default on my fork). That way we can keep the CI runs separate and don't have to roll back if it goes wrong.

Done that in https://github.com/mgorny/pytorch-cpu-feedstock/pull/4. It seems that it created the 16 jobs, but I don't see the CIrun gating job, so I don't think they'll start. But I have another idea: I'll just open another PR to this repo, adding the extra commits, and you'll merge whichever works.

h-vetinari:
> But I have another idea: I'll just open another PR to this repo, adding the extra commits, and you'll merge whichever works.

Sounds good to me!

h-vetinari:
Any comments here @conda-forge/pytorch? CI in #328 (which adds one more change on top of this PR) is looking good and I'd like to merge this soon.

baszalmstra (approved):

jeongseok-meta (approved):

h-vetinari:
Alrighty! Thanks a lot for the reviews! Post-merge comments are still very welcome, we'll have yet another round with #318 coming right up anyway where we can take this up. My goal would be to wrap up that PR in a few days as well and then get 2.6 out the door. 🙃

h-vetinari:
Ah dammit, the workaround for https://github.com/conda-forge/conda-forge-pinning-feedstock/pull/6910, causes windows to fail here 😑 

```
Checking out the ref
  "C:\Program Files\Git\bin\git.exe" checkout --progress --force -B main refs/remotes/origin/main
  Error: error: unable to create file .ci_support/linux_64_blas_implgenericc_compiler_version13channel_targetsconda-forge_maincuda_compilerNonecuda_compiler_versionNonecxx_compiler_version13github_actions_labelscirun-openstack-cpu-2xlargeis_rcFalse.yaml: Filename too long
  Error: error: unable to create file .ci_support/linux_64_blas_implgenericc_compiler_version13channel_targetsconda-forge_maincuda_compilerNonecuda_compiler_versionNonecxx_compiler_version13github_actions_labelscirun-openstack-gpu-2xlargeis_rcFalse.yaml: Filename too long
```

PR #339, opened 2025-01-30
Fix CMake metadata for CUDA-enabled libtorch

h-vetinari:
Follow-up to #318.

Fixes #333 (when tests are passing 🤞)
~Doesn't address #334~ tensorpipe is not supported on windows apparently

PR #342, opened 2025-02-01
 Fix CMake metadata for CUDA-enabled libtorch, take 2

h-vetinari:
This is the same as #339, which is getting close. I want to test the fix for aarch without stopping the current CI run in #339, because I still want to see if the jobs that already ran for ~10h pass (or produce yet another error that needs to fixed).


h-vetinari:
This served its purpose, work was merged through #339

PR #353, opened 2025-02-09
Test skip hyghiene: osx

h-vetinari:
Try to see which skips we still need after 2.6; also, in #326, I've had to skip the `torchinductor` tests because they were segfaulting on osx (in contrast to 2.5), see https://github.com/conda-forge/pytorch-cpu-feedstock/pull/326/commits/b6276415ab18144997288b9d2c47b4c0c3cf28d2. That's not a good thing, so we should ideally fix that.


mgorny:
@h-vetinari, should I include these changes (and the win PR) if I do the next bump?

h-vetinari:
You could use a filtered+squashed version of #352, thanks; this PR isn't ready yet because I started running into #462

PR #356, opened 2025-02-11
[v2.5.x] move `_C.lib` back to `%SP_DIR%\torch\lib`

h-vetinari:
For #354


hmaarrfk:
sorry i missed this, pytorch (and family's) `_C.so` is their python library.

mgorny (approved):
LGTM (provided you remove `temporary` :-))

mgorny:
Oh, yes, that does make sense. We should probably exercise that more often.

hmaarrfk:
> I actually want to merge this as-is, because there's no reason to rebuild the unix variants. 

👀

hmaarrfk:
Lets mark the other windows builds as broken?

Issue #375, opened 2025-03-14
Caffe2/public/cuda.cmake CUDA::nvToolsExt error


das-intensity:
### Solution to issue cannot be found in the documentation.

- [x] I checked the documentation.

### Issue

When attempting to build a cuda version of [momentum-feedstock](https://github.com/jeongseok-meta/momentum-feedstock/tree/linux_64_cuda), we get this error: 
```
2025-02-11T22:19:10.4507601Z   CMake Error at /home/conda/feedstock_root/build_artifacts/momentum_1739311841674/_h_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_pl/share/cmake/Caffe2/public/cuda.cmake:162 (set_property):
2025-02-11T22:19:10.4519742Z     The link interface of target "torch::nvtoolsext" contains:
2025-02-11T22:19:10.4529732Z 
2025-02-11T22:19:10.4540072Z       CUDA::nvToolsExt
2025-02-11T22:19:10.4552673Z 
2025-02-11T22:19:10.4562050Z     but the target was not found.  Possible reasons include:
2025-02-11T22:19:10.4577653Z 
2025-02-11T22:19:10.4595598Z       * There is a typo in the target name.
2025-02-11T22:19:10.4609143Z       * A find_package call is missing for an IMPORTED target.
2025-02-11T22:19:10.4633552Z       * An ALIAS target is missing.
2025-02-11T22:19:10.4645350Z 
2025-02-11T22:19:10.4656027Z   Call Stack (most recent call first):
2025-02-11T22:19:10.4665869Z     /home/conda/feedstock_root/build_artifacts/momentum_1739311841674/_h_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_pl/share/cmake/Caffe2/Caffe2Config.cmake:86 (include)
2025-02-11T22:19:10.4677312Z     /home/conda/feedstock_root/build_artifacts/momentum_1739311841674/_h_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_pl/share/cmake/Torch/TorchConfig.cmake:68 (find_package)
[...]

PR #371, opened 2025-03-11
Pytorch 2.6 + magma 2.9

hmaarrfk:
It is preferring pytorch+cpu+magma 2.9 instead of pytorch + gpu + mamga 2.8

Lets try to see if this compiles
Checklist
* [ ] Used a [personal fork of the feedstock to propose changes](https://conda-forge.org/docs/maintainer/updating_pkgs.html#forking-and-pull-requests)
* [ ] Bumped the build number (if the version is unchanged)
* [ ] Reset the build number to `0` (if the version changed)
* [ ] [Re-rendered]( https://conda-forge.org/docs/maintainer/updating_pkgs.html#rerendering-feedstocks ) with the latest `conda-smithy` (Use the phrase <code>@<space/>conda-forge-admin, please rerender</code> in a comment in this PR for automated rerendering)
* [ ] Ensured the license file is being packaged.

Issue #357, opened 2025-02-12
CMake metadata for CUDA builds not yet self-contained


h-vetinari:
I'm seeing the following in a CMake-based feedstock
```
  CMake Error at $PREFIX/share/cmake/Caffe2/public/cuda.cmake:162 (set_property):
    The link interface of target "torch::nvtoolsext" contains:

      CUDA::nvToolsExt

    but the target was not found.  Possible reasons include:

      * There is a typo in the target name.
      * A find_package call is missing for an IMPORTED target.
      * An ALIAS target is missing.

  Call Stack (most recent call first):
    $PREFIX/share/cmake/Caffe2/Caffe2Config.cmake:86 (include)
    $PREFIX/lib/python3.10/site-packages/torch/share/cmake/Torch/TorchConfig.cmake:68 (find_package)
    CMakeLists.txt:79 (find_package)
```

The upstream [sources](https://github.com/pytorch/pytorch/blob/v2.5.1/cmake/public/cuda.cmake#L172-L183) for this are affected by
https://github.com/conda-forge/pytorch-cpu-feedstock/blob/be203905856551e5a4b50dd03c6be2e9ddef5fc4/recipe/patches/0003-Add-USE_SYSTEM_NVTX-option-138287.patch#L50-L58
and end up looking like
```CMake
# nvToolsExt
if(USE_SYSTEM_NVTX)
  find_path(nvtx3_dir NAMES nvtx3)
else()
  find_path(nvtx3_dir NAMES nvtx3 PATHS "${PROJECT_SOURCE_DIR}/third_party/NVTX/c/include" NO_DEFAULT_PATH)
endif()
find_package_handle_standard_args(nvtx3 DEFAULT_MSG nvtx3_dir)
if(nvtx3_FOUND)
  add_library(torch::nvtx3 INTERFACE IMPORTED)
  target_include_directories(torch::nvtx3 INTERFACE "${nvtx3_dir}")
  target_compile_definitions(torch::nvtx3 INTERFACE TORCH_CUDA_USE_NVTX3)
else()
  message(WARNING "Cannot find NVTX3, find old NVTX instead")
  add_library(torch::nvtoolsext INTERFACE IMPORTED)
  set_property(TARGET torch::nvtoolsext PROPERTY INTERFACE_LINK_LIBRARIES CUDA::nvToolsExt)
endif()
```

[...]

PR #377, opened 2025-03-18
Fix missing symbols of pytorch_python on Windows

jeongseok-meta:
Checklist


jeongseok-meta:

jeongseok-meta:
Do we want to update the cmake test to test the effect of this patch? It's currently only test if the packages can be found, but we may need to additionally test building a simple program to see if the linker error is fixed. For example, we could use a test similar to https://github.com/jeongseok-meta/libtorch-example/blob/main/main.cpp

hmaarrfk:
> Do we want to update the [cmake test](https://github.com/conda-forge/pytorch-cpu-feedstock/blob/main/recipe/cmake_test/CMakeLists.txt) to test the effect of this patch?

as you wish.

jeongseok-meta:
Added cuda-nvtx-dev for cmake_test to resolve:

```
2025-03-19T09:18:56.0414340Z CMake Error at /home/conda/feedstock_root/build_artifacts/libtorch_1742344247475/_test_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placeho/share/cmake/Caffe2/public/cuda.cmake:162 (set_property):
2025-03-19T09:18:56.0417934Z   The link interface of target "torch::nvtoolsext" contains:
2025-03-19T09:18:56.0418643Z 
2025-03-19T09:18:56.0418872Z     CUDA::nvToolsExt
2025-03-19T09:18:56.0419240Z 
2025-03-19T09:18:56.0419631Z   but the target was not found.  Possible reasons include:
2025-03-19T09:18:56.0420255Z 
2025-03-19T09:18:56.0420525Z     * There is a typo in the target name.
2025-03-19T09:18:56.0421396Z     * A find_package call is missing for an IMPORTED target.
2025-03-19T09:18:56.0422754Z     * An ALIAS target is missing.
```

as reported in https://github.com/conda-forge/pytorch-cpu-feedstock/issues/375

isuruf:
> I would rather add this so the run-requirements of libtorch.

I'd rather not. We don't add any development dependencies like `mkl-devel` to run requirements.

h-vetinari:
> We don't add any development dependencies like `mkl-devel` to run requirements.

It's required for CMake to run `find_package(torch)` for CUDA-builds, see #375. For our downstreams, we really should have working CMake metadata out of the box (our testing here currently hides the problem through adding extra test dependencies).

I agree that run-requiring dev dependencies is unfortunate, but that's what pytorch currently does. You know the upstream repo pretty well, perhaps you want to propose a patch that allows CMake to find torch without `cuda-nvtx-dev` present.

isuruf:
Let's not do two things in this PR and discuss in the other issue.

h-vetinari:
@jeongseok-meta, this is not working yet

```
CMake Error at CMakeLists.txt:12 (find_library):
  Could not find torch_python using the following names: torch_python
```

While you're debugging, please limit the number of build configs that are necessary to 1-2 on linux. I've cancelled the builds now to merge the 2.4 kineto PR.

h-vetinari:
> Let's not do two things in this PR and discuss in the other issue.

I'd prefer to fix this for users now, but OK to revert https://github.com/conda-forge/pytorch-cpu-feedstock/pull/377/commits/7484a4a3d6d762c4b8eca16c12e31ca7c65e5c0c as long as we fix #375 reasonably soon, one way or another.

jeongseok-meta:

jeongseok-meta:

isuruf:
Let's revert https://github.com/conda-forge/pytorch-cpu-feedstock/commit/7484a4a3d6d762c4b8eca16c12e31ca7c65e5c0c as that's not the correct fix and this only affects downstream users who build extensions against torch, which are rare and they can workaround by adding `cuda-nvtx-dev` themselves.

jeongseok-meta:
> this is not working yet

It appears that the test fails because the `torch_python` shared library is copied when building `pytorch`, but not yet for `libtorch`. This issue can be seen in the following code snippet:

https://github.com/conda-forge/pytorch-cpu-feedstock/blob/f3f31a9ee57ab64d53cbc171502ed477e3ac76df/recipe/build.sh#L281C10-L281C16

To resolve this, we have three potential options:

(A) Implement a workaround by finding the shared library from its path before it is copied. Although this approach may work, it requires thorough testing to ensure compatibility with downstream processes.
(B) Copy the file when building libtorch. However, it is uncertain whether this solution will be effective.
(C) Split the test, as the torch_python test is currently only executed for pytorch and not for libtorch.

Please let me know which option you prefer or if there's a better solution.

jeongseok-meta:
> Let's revert [7484a4a](https://github.com/conda-forge/pytorch-cpu-feedstock/commit/7484a4a3d6d762c4b8eca16c12e31ca7c65e5c0c) as that's not the correct fix and this only affects downstream users who build extensions against torch, which are rare and they can workaround by adding `cuda-nvtx-dev` themselves.

Thanks for the suggestion. I'm open to either option and would like to leave the decision up to a consensus with other maintainers, as you guys may have a better understanding of the conda-forge (and/or this feedstock) convention or philosophy. I'm flexible and willing to go with the consensus.

recipe/meta.yaml
```diff
@@ -238,13 +236,6 @@ test:
     - test -f $PREFIX/share/cmake/Torch/TorchConfig.cmake                       # [linux]
     - if not exist %LIBRARY_PREFIX%\share\cmake\Torch\TorchConfig.cmake exit 1  # [win]
 
-    # test integrity of CMake metadata
-    - cd cmake_test
-    - cmake -GNinja -DCMAKE_CXX_STANDARD=17 $CMAKE_ARGS .   # [unix]
-    - cmake -GNinja -DCMAKE_CXX_STANDARD=17 %CMAKE_ARGS% .  # [win]
```
h-vetinari, line 238: Don't move the entire `cmake_test`. The config stage still needs to be tested in `libtorch`.

recipe/meta.yaml
```diff
@@ -205,8 +210,7 @@ test:
     - cmake
     - ninja
     - pkg-config
-  files:
-    - cmake_test/
+    - pybind11
```
h-vetinari, line 213: also, this test dependency needs to move to pytorch as well. At this point, please rebase this PR to reduce the commit churn a bit (you can also leave out 7484a4a3d6d762c4b8eca16c12e31ca7c65e5c0c right away rather than reverting it).

jeongseok-meta:

recipe/meta.yaml
h-vetinari: > also, this test dependency needs to move to pytorch as well.

Thanks for the rebase. What I meant is that `pybind11` should actually move, i.e. not be a test dependency of `libtorch` anymore. The rest looks OK

recipe/meta.yaml
```diff
@@ -507,6 +517,13 @@ outputs:
         - test ! -f $SP_DIR/functorch/__pycache__/__init__.cpython-313.pyc          # [py!=313 and unix]
         - if exist %SP_DIR%\functorch\__pycache__\__init__.cpython-313.pyc exit 1   # [py!=313 and win]
 
+        # test integrity of CMake metadata
```
h-vetinari, line 521: This comment could be extended, like "and that <...> is visible as a symbol from libtorch_python".

recipe/meta.yaml
h-vetinari: Sorry, I resolved too quickly here, this is still not done.

jeongseok-meta: Hmm, don't we need it for both tests? Both tests use the same `CMakeLists.txt` (`libtorch` just to test CMake configuration, while `pytorch` tests both CMake configuration and building/linking), which requires `pybind11` as per the `find_package(pybind11)` in the `CMakeLists.txt`.

h-vetinari: Ah, right. That's because we're not using separate CMake tests for the two outputs. In this case, this is fine.

jeongseok-meta: Thanks for confirming!

recipe/patches/0017-Hide-torch_python-symbols-142214.patch
```diff
@@ -0,0 +1,268 @@
...
+diff --git a/torch/CMakeLists.txt b/torch/CMakeLists.txt
+index 650319cb1ee..aebd0d1f442 100644
+--- a/torch/CMakeLists.txt
++++ b/torch/CMakeLists.txt
+@@ -310,7 +310,7 @@ endif()
+ 
+ add_library(torch_python SHARED ${TORCH_PYTHON_SRCS})
+ torch_compile_options(torch_python)  # see cmake/public/utils.cmake
+-if(NOT WIN32)
++if(APPLE)
```
isuruf, line 33: Let's keep this `NOT WIN32`. Otherwise, it's a ABI breaking change.

jeongseok-meta:
It appears that the Windows CI jobs have not been triggered. Do I require additional permissions to initiate them? If so, could someone please assist me in triggering the CI or provide guidance on how to obtain the necessary permissions?

Tobias-Fischer:
See e.g. https://github.com/conda-forge/.cirun/commit/0d90fff8e701713463942d7af85274046eee18a2

jeongseok-meta:
> See e.g. [conda-forge/.cirun@0d90fff](https://github.com/conda-forge/.cirun/commit/0d90fff8e701713463942d7af85274046eee18a2)

Submitted a PR. Thanks!

https://github.com/conda-forge/.cirun/pull/33

jeongseok-meta:
`linux_64_blas_implgenericc_compiler_version13channel_targetsconda-forge_maincuda_compilercuda-nvcccuda_compiler_version12.6cxx_compiler_version13is_rcFalse` still fails, even with `cuda-nvtx-dev`:

```
2025-03-21T05:14:16.9515625Z CMake Error at /home/conda/feedstock_root/build_artifacts/libtorch_1742501186005/_test_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placeho/share/cmake/Caffe2/public/cuda.cmake:162 (set_property):
2025-03-21T05:14:16.9519698Z   The link interface of target "torch::nvtoolsext" contains:
2025-03-21T05:14:16.9520358Z 
2025-03-21T05:14:16.9520595Z     CUDA::nvToolsExt
2025-03-21T05:14:16.9520951Z 
2025-03-21T05:14:16.9521348Z   but the target was not found.  Possible reasons include:
2025-03-21T05:14:16.9521994Z 
2025-03-21T05:14:16.9522800Z     * There is a typo in the target name.
2025-03-21T05:14:16.9523700Z     * A find_package call is missing for an IMPORTED target.
2025-03-21T05:14:16.9524590Z     * An ALIAS target is missing.
2025-03-21T05:14:16.9525039Z 
2025-03-21T05:14:16.9525452Z Call Stack (most recent call first):
2025-03-21T05:14:16.9528605Z   /home/conda/feedstock_root/build_artifacts/libtorch_1742501186005/_test_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placeho/share/cmake/Caffe2/Caffe2Config.cmake:86 (include)
2025-03-21T05:14:16.9534174Z   /home/conda/feedstock_root/build_artifacts/libtorch_1742501186005/_test_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placeho/share/cmake/Torch/TorchConfig.cmake:68 (find_package)
2025-03-21T05:14:16.9537347Z   CMakeLists.txt:8 (find_package)
```

jeongseok-meta:
It turned out `cuda-nvtx-dev` wasn't added for the cmake_test for `pytorch`. Fixed in [ddaf0dd](https://github.com/conda-forge/pytorch-cpu-feedstock/pull/377/commits/ddaf0dd600e5b65966745fcd90cc6a1e0353d327).

jeongseok-meta:

recipe/cmake_test/CMakeLists.txt
```diff
@@ -1,4 +1,45 @@
 project(cf_dummy LANGUAGES C CXX)
 cmake_minimum_required(VERSION 3.12)
+
+option(WITH_TORCH_PYTHON "Build with torch_python" OFF)
+
+find_package(Python3 REQUIRED COMPONENTS Interpreter Development)
+find_package(pybind11 CONFIG REQUIRED)
```
h-vetinari, line 7: Can you move this behind `if(WITH_TORCH_PYTHON)` please? That way we can leave out `pybind11` as a test dependency for `libtorch`.

recipe/cmake_test/CMakeLists.txt
```diff
@@ -1,4 +1,45 @@
...
+if(WITH_TORCH_PYTHON)
+  if(WIN32)
+    set(libtorch_base_path $ENV{PREFIX}/Lib/site-packages/torch)
+  else()
+    set(libtorch_base_path $ENV{PREFIX})
+  endif()
+  find_library(torch_python NAMES torch_python HINTS ${libtorch_base_path}/lib/ REQUIRED)
+endif()
+
+set(target_name cmake_test)
```
h-vetinari, line 20: as long as we only have one target, adding a variable for it is not helpful. Please directly use `cmake_test` below

recipe/cmake_test/CMakeLists.txt
```diff
@@ -1,4 +1,45 @@
...
+
+if(WITH_TORCH_PYTHON)
+  target_link_libraries(${target_name} PRIVATE
+    ${torch_python}
+  )
+endif()
+
+target_compile_options(${target_name} PRIVATE
+  ${TORCH_CXX_FLAGS}
+)
```
h-vetinari, line 45: I think basically everything you're adding here can go into one big `if(WITH_TORCH_PYTHON)` branch (the one at line 11 currently).

recipe/meta.yaml
h-vetinari: With the CMake file adapted correspondingly, this should then now be removed for real

jeongseok-meta:
The cmake_test still fails on Windows with the following errors:

```
  CMake Error at $PREFIX/share/cmake/Caffe2/public/cuda.cmake:162 (set_property):
    The link interface of target "torch::nvtoolsext" contains:

      CUDA::nvToolsExt

    but the target was not found.  Possible reasons include:

      * There is a typo in the target name.
      * A find_package call is missing for an IMPORTED target.
      * An ALIAS target is missing.

  Call Stack (most recent call first):
```

which is also reported in #357 and #375.

To address this error, I'd like to test a new hypothesis. We utilize system nvtx by setting the `USE_SYSTEM_NVTX` option to `1`, instead of relying on the submodule from the upstream PyTorch repository:

https://github.com/conda-forge/pytorch-cpu-feedstock/blob/f3f31a9ee57ab64d53cbc171502ed477e3ac76df/recipe/build.sh#L224

The issue arises because the `cuda.cmake` file is installed as-is and still [relies on the `USE_SYSTEM_NVTX` option to locate nvtx](https://github.com/pytorch/pytorch/blob/v2.6.0/cmake/public/cuda.cmake#L172-L187) when called by downstream `find_package(Torch)`:

```cmake
# nvToolsExt
if(USE_SYSTEM_NVTX)
  find_path(nvtx3_dir NAMES nvtx3)
else()
  find_path(nvtx3_dir NAMES nvtx3 PATHS "${PROJECT_SOURCE_DIR}/third_party/NVTX/c/include" NO_DEFAULT_PATH)
endif()
find_package_handle_standard_args(nvtx3 DEFAULT_MSG nvtx3_dir)
if(nvtx3_FOUND)
  add_library(torch::nvtx3 INTERFACE IMPORTED)
  target_include_directories(torch::nvtx3 INTERFACE "${nvtx3_dir}")
  target_compile_definitions(torch::nvtx3 INTERFACE TORCH_CUDA_USE_NVTX3)
else()
  message(WARNING "Cannot find NVTX3, find old NVTX instead")
  add_library(torch::nvtoolsext INTERFACE IMPORTED)
  set_property(TARGET torch::nvtoolsext PROPERTY INTERFACE_LINK_LIBRARIES CUDA::nvToolsExt)
endif()
```

Unless the downstream project sets this option, which is not ideal, it attempts to find nvtx in the PyTorch third-party subdirectory. Since this path is invalid, it defaults to using `CUDA::nvToolsExt`, which is not available (the reason for this discrepancy with libtorch's built-in nvtx3 is unclear). As a potential solution, I'm trying out a patch that ensures `cuda.cmake` always uses system nvtx.

To make this work, `nvtx-c` is added as a test dependency. This suggests we may want to create a `libtorch-dev` output including `libtorch` and dev dependencies like `pybind11` and `nvtx-c` to make it self-contained.

jeongseok-meta:
The Linux build is still running, but I have locally confirmed that it passes. That said, it's ready for review.

recipe/patches/0018-Always-use-system-nvtx.patch
```diff
@@ -0,0 +1,25 @@
...
+
+diff --git a/cmake/public/cuda.cmake b/cmake/public/cuda.cmake
+index 792e2ac78d3..cb1ece22cf3 100644
+--- a/cmake/public/cuda.cmake
++++ b/cmake/public/cuda.cmake
+@@ -146,7 +146,7 @@ else()
+ endif()
+ 
+ # nvToolsExt
+-if(USE_SYSTEM_NVTX)
```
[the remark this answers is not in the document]
jeongseok-meta: Yeah, I think it's something to discuss upstream. To me, it should be configurable via options (like creating `cuda.cmake.in`). I'll submit a PR for this and let's see how it goes!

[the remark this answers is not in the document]
jeongseok-meta: https://github.com/pytorch/pytorch/pull/149861

recipe/meta.yaml
```diff
@@ -386,7 +386,7 @@ outputs:
         # for cmake_test
         - cmake
         - cuda-nvrtc-dev            # [cuda_compiler_version != "None"]
-        - ç             # [cuda_compiler_version != "None"]
+        - cuda-nvtx-dev             # [cuda_compiler_version != "None"]
```
h-vetinari: like this one

h-vetinari (changes requested):
Thank you for your patience here, and for following up upstream! The PR looks good to me, aside from the fact that it has way too much commit churn (especially around nvtx) going back and forth between different things. I've done a test rebase and could reduce this PR to four well-separated commits (all with your authorship of course).

Since #371 is also ready, this is a good chance to do a joint merge, which I'll open shortly.

Thanks a lot for the PR! 🙏

PR #391, opened 2025-05-29
pytorch 2.7.1; switch label on windows; turn on artefact persistence

h-vetinari:
These are each intended as a work-around for https://github.com/conda/infrastructure/issues/1159. Let's hope at least one works out.

PR #395, opened 2025-07-03
Fix artefact persistence on windows

h-vetinari:
In #391 we tried to add work-arounds for https://github.com/conda/infrastructure/issues/1159; however, the artefact generation step is not yet functional on windows for GHA, see https://github.com/conda-forge/conda-smithy/issues/2345.

After merging #393, things failed again on main, for win+CUDA and aarch+CUDA. This shows a second problem that can be quite annoying: the artefact generation is currently being run unconditionally, even when builds are cancelled. Over an hour after the job has been cancelled by timeout, the artefact generation (which is completely unnecessary in this case) has still not finished. Try to address this as well.

Actual fixes need to happen in smithy; this is both for developing the fix, as well as getting windows builds up.

### DIFF
--- a/.github/workflows/conda-build.yml
+++ b/.github/workflows/conda-build.yml
@@ -213,7 +213,7 @@ jobs:
           export BLD_ARTIFACT_PREFIX="conda_pkgs"
         fi
         if [ $OS == "windows" ]; then
-          cmd.exe /c .scripts\\create_conda_build_artifacts.bat
+          pwsh -Command ". '.scripts/create_conda_build_artifacts.bat'"
         else
           ./.scripts/create_conda_build_artifacts.sh
         fi


h-vetinari:
> Do I understand correctly that these changes will keep being overwritten on rerender?

See OP:

> Actual fixes need to happen in smithy; this is both for developing the fix, as well as getting windows builds up.

:)

mgorny:
Thanks for doing this!

h-vetinari:
OK, I finally have the artefact creating working - [here](https://github.com/conda-forge/pytorch-cpu-feedstock/actions/runs/16084100609/attempts/1) it is for eb308200e690e64f9e64cf628566e87e0f94ad98.

This is a bit of a special case not really considered by CFEP-3; the build has been fully produced by CI, with publicly visible logs, so the usual process of uploading to a private channel seems unnecessary? In any case, I did push them to my private channel, and have now run
```
DELEGATE=h-vetinari
PACKAGE_VERSION=2.7.1
for package in libtorch pytorch pytorch-gpu; do
  anaconda copy --from-label main --to-label main --to-owner conda-forge ${DELEGATE}/${package}/${PACKAGE_VERSION}
done
```

PR #407, opened 2025-08-01
Fix Intel/LLVM openmp build

isuruf:
Checklist
* [ ] Used a [personal fork of the feedstock to propose changes](https://conda-forge.org/docs/maintainer/updating_pkgs.html#forking-and-pull-requests)
* [ ] Bumped the build number (if the version is unchanged)
* [ ] Reset the build number to `0` (if the version changed)
* [ ] [Re-rendered]( https://conda-forge.org/docs/maintainer/updating_pkgs.html#rerendering-feedstocks ) with the latest `conda-smithy` (Use the phrase <code>@<space/>conda-forge-admin, please rerender</code> in a comment in this PR for automated rerendering)
* [ ] Ensured the license file is being packaged.

PR #409, opened 2025-08-07
V2.8.0

mgorny:
Checklist


h-vetinari:
The [release notes](https://github.com/pytorch/pytorch/releases/tag/v2.8.0) make it sound like we'll need to double-check nvtx support as well

> A downstream project using `-DUSE_SYSTEM_NVTX` will not be able to find NVTX3 or `torch::nvtx3` via PyTorch's `cmake/public/cuda.cmake`. The downstream project now needs to explicitly find NVTX3 and `torch::nvtx3` by implementing the same logic in PyTorch's `cmake/Dependences.cmake`.

mgorny:
> The [release notes](https://github.com/pytorch/pytorch/releases/tag/v2.8.0) make it sound like we'll need to double-check nvtx support as well
> 
> > A downstream project using `-DUSE_SYSTEM_NVTX` will not be able to find NVTX3 or `torch::nvtx3` via PyTorch's `cmake/public/cuda.cmake`. The downstream project now needs to explicitly find NVTX3 and `torch::nvtx3` by implementing the same logic in PyTorch's `cmake/Dependences.cmake`.

From what I understand, this means checking reverse dependencies.

mgorny:
Wait, I'm reading the output wrong. Investigating further.

mgorny:
Okay, I've learned more about Windows shell than I wanted to know, and I suspect delayed expansion did no work as expected. I've tried replacing `%PY_VERSION_FULL%` with `!PY_VERSION_FULL!`, and apparently this one works at least in Wine's implementation of `cmd`.

h-vetinari:
Haha, the win+CUDA build manages to blow through the disk space
```
error: [Errno 28] No space left on device
```
despite using the largest runner alread
https://github.com/conda-forge/pytorch-cpu-feedstock/blob/d21fc6dbb4db827872fc8cf05e70f6771bd4937b/recipe/conda_build_config.yaml#L31

But perhaps the "size" of that runner is only measured in CPUs/RAM, not storage? Could we extend that @aktech @wolfv?

mgorny:
Well, at least the Python version issue was fixed. Also, looks like my idea of using `pytest --forked` won't work for CUDA tests.

mgorny:
Le sigh, I've fetched the artifact and couldn't reproduce the segfaults locally. But I've noticed that our openblas+openmp constraint didn't work anymore. Let's try again.

By the way, I'm wondering if we should perhaps skip CPU tests in CUDA builds. I see they're some of the longest tests in CI, and I suppose it's sufficient that we test them in CPU builds.

h-vetinari:
> But I've noticed that our openblas+openmp constraint didn't work anymore.

Probably related to #407, which unfortunately isn't green either.

mgorny:
Uh, so I guess CMake 4 breaks AArch64 builds? I'll try debugging that locally.

mgorny:
> I thought we had a workaround with a comment in our script for this

Ah, sorry, indeed, I was doing a non-CUDA build and didn't notice it's there for CUDA.

aktech:
> @aktech @wolfv, could we increase disk space on the windows agents?

https://docs.cirun.io/reference/yaml#custom-disk-size-for-azure

The [windows runners configuration](https://github.com/conda-forge/.cirun/blob/6b7b5ddc2e8a47125cbe63f1cb5c8ca2566ddec6/.cirun.global.yml#L121) needs to be updated with: 

```yml
    extra_config:
      storageProfile:
        osDisk:
          diskSizeGB: 512
```

h-vetinari:
> We're seeing a humongous regression in disk space use (from ~40G to over 90G).

Perhaps it's time to give the v1 conversion another shot. The collaboration/interaction with anaconda didn't materialise (aside from the initial push with @danpetry at the beginning of the year), so the ability to merge back changes from anaconda shouldn't be a blocker anymore.

Speaking of the anaconda recipe, weirdly, https://github.com/AnacondaRecipes/pytorch-feedstock/blob/master/recipe/meta.yaml is still on 2.6, but anaconda has 2.8 [builds](https://anaconda.org/conda-forge/libtorch/files) already, so it seems the feedstock doesn't reflect everything.

CC @JamesRobertsonGames @jezdez

mgorny:
Yeah, I can do that, but I'm not sure if it will suffice, given we're pretty much close to running out of space in a single build phase.

That said, 90G free sounds suspiciously low for a 500G volume too. Maybe we need some other knob to make it work right?

rgommers:
> Perhaps it's time to give the v1 conversion another shot. 

+1 that makes a lot of sense. It's a major struggle almost every time to get a new release out, and the large improvement in both performance and debuggability (is that a word?) seems very much necessary.

mgorny:
> Unrelated to the storage question, but still relevant for this PR: it looks like we also have to incorporate #413 here.

Indeed. Seems like the conclusion is to use vendored pybind11, do I read that correctly?

That said, given the Windows problem, I'm wondering if I should try another PR with all the pending changes applied on top of 2.7.1, and possibly try v1 recipe there as well (though I'd do that separately). Having https://github.com/conda/ceps/pull/110 standardized would also be helpful for v1 recipes.

aktech:
>@aktech, could you check if the changes took effect?

The default disk on the VM is **600 GB**, which is a **temporary storage disk**.  
This cannot be increased unless we bump to `Standard_D32ads_v5`:

- [Azure VM dadsv5-series sizes](https://learn.microsoft.com/en-us/azure/virtual-machines/sizes/general-purpose/dadsv5-series?utm_source=chatgpt.com&tabs=sizestoragelocal#sizes-in-series)  
- [Temporary disk overview](https://learn.microsoft.com/en-us/azure/virtual-machines/managed-disks-overview#temporary-disk)

The **1000 GB** we added is a **managed unallocated disk** — you’ll need to switch to using that instead of relying on the default 600 GB temp disk.

Here is a test workflow example showing the 1000GB disk: https://github.com/aktech/cirun-azure-resize-disk/actions/runs/17826848401/job/50681943221#step:4:35

TLDR; if you don't wanna do the resizing / (allocating the managed disk) in CI, we could use [Standard_D32ads_v5](https://learn.microsoft.com/en-us/azure/virtual-machines/sizes/general-purpose/dadsv5-series?utm_source=chatgpt.com&tabs=sizestoragelocal#sizes-in-series) which has 1200GB of temporary disk by default.

mgorny:
Hmm, perhaps a good first step would be to build on `D:\` instead of `C:\`? Though FWICS it isn't configurable right now, so we'd have to keep editing the workflow after every rerender.

@aktech, am I thinking correctly that we could do the resizing from `build.bat`? Though ofc it kinda sucks to have such host-local overrides in the recipe.

Lemme try overriding the build to use `D:\` for the time being.

mgorny:
Unfortunately, this change didn't seem to actually work:

```
  ==> envvars <==
  allow_softlinks: False
  solver: libmamba
  bld_path: C:\\bld\\
```

.github/workflows/conda-build.yml
```diff
@@ -133,6 +133,31 @@ jobs:
         fi
         ./.scripts/run_osx_build.sh
 
+    # https://github.com/aktech/cirun-azure-resize-disk
+    - name: Resize all partitions to maximum
+      if: matrix.os == 'windows'
+      shell: pwsh
```
h-vetinari, line 139: This will need to move to smithy (probably with a config in `conda-forge.yml`), otherwise it'll get deleted on every rerender.

mgorny: https://github.com/conda-forge/conda-smithy/issues/2394

mgorny:
Uh, I accidentally rerendered over the Windows fix :facepalm:.

Let me read #413 up, in case I should change something before restarting.

h-vetinari (approved):
Thanks a lot for the persistence on this one! ❤️ 

Discussion about the pybind situation in #413 is still ongoing with the pybind maintainers, so a rebuild for v3 (or removing the dependence on `pybind-abi` completely) can be done in a follow-up.

mgorny:
> Thanks a lot for the persistence on this one! ❤️

No problem. I'm sorry it took this long — I have made more mistakes than i should have, notably failed to pin `run` dependencies early on, which would have saved me a lot of subsequent testing.

There's also the open question on how to deal with cudnn. I'm not even sure if this is something to report to PyTorch or to NVIDIA.

h-vetinari (commented):
> There's also the open question on how to deal with cudnn. I'm not even sure if this is something to report to PyTorch or to NVIDIA.

I think we could start at least with an issue on the cudnn feedstock, at least write down the things you remember from that debugging session somewhere, before it becomes just a haze. 😅

h-vetinari:
Windows CUDA builds are still failing to upload, done manually from the artefacts:
```console
$ gh run download 17899219922 --repo conda-forge/pytorch-cpu-feedstock --name conda_artifacts_17899219922_win_64_channel_targetsconda-forge_maincu_hca575dce
$ unzip pytorch-cpu-feedstock_conda_artifacts_.zip
$ cd bld/win-64
$ rm current_repodata.json index.html repodata*
$ ls
libtorch-2.8.0-cuda128_mkl_ha34d6f4_300.conda       pytorch-2.8.0-cuda128_mkl_py312_h0850830_300.conda
pytorch-2.8.0-cuda128_mkl_py310_h0b8c608_300.conda  pytorch-2.8.0-cuda128_mkl_py313_hf206996_300.conda
pytorch-2.8.0-cuda128_mkl_py311_hd9a8a8a_300.conda  pytorch-gpu-2.8.0-cuda128_mkl_h2fd0c33_300.conda
$ ls | xargs anaconda upload
$ DELEGATE=h-vetinari
PACKAGE_VERSION=2.8.0
for package in libtorch pytorch pytorch-gpu; do
  anaconda copy --from-label main --to-label main --to-owner conda-forge ${DELEGATE}/${package}/${PACKAGE_VERSION}
done
```

mgorny:
Presumably we'll need to restart that one failed AArch64 build — but I don't see the rerun button right now, so I guess it'll only appear when the other job is finished. Not sure if I can rerun it without rerunning the Windows build though.

PR #445, opened 2025-11-02
[DO NOT MERGE] Debug upload

h-vetinari:
For https://github.com/conda/infrastructure/issues/1159


h-vetinari:
Some notes for myself:
* there's no tcpdump on windows (though that might _hopefully_ change at some point, see https://github.com/nmap/npcap/issues/73)
* windump and npcap are not easily available (because unlisted) on chocolatey; exact pins didn't help
* wireshark can be installed by choco, but doesn't appear to provide the `tshark` CLI; various attempts at fixing this were not effective
* trying to install wireshark from the base installer following the [docs](https://www.wireshark.org/docs/wsug_html_chunked/ChBuildInstallWinInstall.html) is severely limited; it seems to be wholly unsuitable for CLI usage (and if using the CLI-friendly `/S` silent option, the npcap library we need won't get installed)
* installer seems to hang in CLI
* ... rest TBD

h-vetinari:
This is obsolete now with the switch to namespace runners (c.f. #524)

PR #484, opened 2026-02-06
(DNM) Demo cirrus runners

jaimergp:
Checklist


jaimergp:
One thing to note is that in the case of Cirrus Runners, and unlike Cirun, the runner label doubles as the (versioned!) runner image (e.g. `ghcr.io/cirruslabs/ubuntu-runner-amd64-gpu:24.04`).

This not necessarily a bad thing, but it will prevent us from performing VM image bumps via conda-smithy rerenders. So I'd suggest we recommend unversioned labels which we then finalize during rerender using default values chosen in conda-smithy.

jaimergp:
Cirrus (and hence Runners) has been [acquired](https://cirruslabs.org) by OpenAI and will shutdown on June 1st., so I won't investigate any further

PR #495, opened 2026-03-30
[skip ci] move to activation json

wolfv:
The activation scripts were pretty weird and broken.

The "backup" was never useful (we never overwrote the flags anyways).

The `sed` command didn't work on Windows.

This is an attempt to use the `env_vars.d` JSON files instead and write the values properly using a little Python helper script.


hmaarrfk:
don't you want to bump the build number?

ruben-arts:
What about merging it into #496? Then we can skip having two builds that solve related issues.

PR #496, opened 2026-03-31
misc: remove activation scripts and replace with json files

ruben-arts:
Checklist
* [x] Used a [personal fork of the feedstock to propose changes](https://conda-forge.org/docs/maintainer/updating_pkgs.html#forking-and-pull-requests)
* [na] Bumped the build number: **Didn't update build-number as this change doesn't need a dedicated rebuild.**
* [na] Reset the build number to `0` (if the version changed)
* [na] [Re-rendered]( https://conda-forge.org/docs/maintainer/updating_pkgs.html#rerendering-feedstocks ) with the latest `conda-smithy` (Use the phrase <code>@<space/>conda-forge-admin, please rerender</code> in a comment in this PR for automated rerendering)
* [na] Ensured the license file is being packaged.

---

This uses the `env_vars.d` JSON files instead and write the values properly using a little Python helper script.

Removing the activations scripts and having a cross platform script that's tested on all platforms.

PR #490, opened 2026-02-23
v2.11.0

mgorny:
Checklist
* [x] Used a [personal fork of the feedstock to propose changes](https://conda-forge.org/docs/maintainer/updating_pkgs.html#forking-and-pull-requests)
* [ ] Bumped the build number (if the version is unchanged)
* [x] Reset the build number to `0` (if the version changed)
* [x] [Re-rendered]( https://conda-forge.org/docs/maintainer/updating_pkgs.html#rerendering-feedstocks ) with the latest `conda-smithy` (Use the phrase <code>@<space/>conda-forge-admin, please rerender</code> in a comment in this PR for automated rerendering)
* [x] Ensured the license file is being packaged.

PR #508, opened 2026-06-05
backport patch for GCC 14 compat

### LINKED ISSUE #494: Using the 10.1 CUDA arch doesn't work anymore.
### Comment:

Hi all, I'm try to build the `torchvision` package for a Tegra device. After intially failing I'm trying again by using the `CF_TORCH_ARCH_LIST` set by the activation script of `pytorch`. See PR: https://github.com/conda-forge/torchvision-feedstock/pull/155

This fails due to the build not finding support for the `10.1+PTX` version which is set by the build script of `pytorch`: https://github.com/conda-forge/pytorch-cpu-feedstock/blob/cdeaf5ae3973f89d8e758ff05a469441a29b1722/recipe/build.sh#L234

In the pytorch code this version was redacted and replaced by `11.0`: https://github.com/pytorch/pytorch/commit/9c5601ecc316e5be548038bc24411acd7c74a032

I think this should also be done in the buildscript here. I'm surprised that it even works as this was already an old change 🤔.

TL;DR: can we change `TORCH_CUDA_ARCH_LIST="8.7;10.1+PTX"` to `TORCH_CUDA_ARCH_LIST="8.7;11.0+PTX"`?

fyi @traversaro

### LINKED ISSUE #507: Incorrect runtime error about GCC 14 + CUDA 12.9
See
```
    File "$PREFIX/lib/python3.11/site-packages/torch/utils/cpp_extension.py", line 716, in build_extensions
      _check_cuda_version(compiler_name, compiler_version)
    File "$PREFIX/lib/python3.11/site-packages/torch/utils/cpp_extension.py", line 574, in _check_cuda_version
      raise RuntimeError(
  RuntimeError: The current installed version of $BUILD_PREFIX/bin/x86_64-conda-linux-gnu-c++ (>=6.0.0, <14.0).
```

We need to backport https://github.com/pytorch/pytorch/commit/39565a7dcf8f93ea22cedeaa20088b24ff6d2634

h-vetinari:
Fixes #507 
Fixes #494

While we're at it, go native for aarch64 builds.

This includes https://github.com/conda-forge/pytorch-cpu-feedstock/commit/7cdb3ec691ce62f8af1d19cbfecba2b43302b0c9 from #496 ~and https://github.com/conda-forge/pytorch-cpu-feedstock/commit/641ed7ef33a9b9d3b0c6229fde59b4513791e6bc that was rebased out of #490~ (nope, still errors)


hmaarrfk:
Nice

mgorny:
> While we're at it, go native for aarch64 builds.

I've reverted going native because I couldn't get any of the builders to reliably survive building it. And yes, I've tried a lot. Maybe the Depot builders will be stronger than that, but at this point, I don't think switching is worth the effort, not to mention how much slower they are.

h-vetinari:
Yeah, I came to the same conclusion now. We can try the depot builders later. For now I've gone back to the status quo ante.

recipe/meta.yaml
```diff
@@ -548,41 +575,19 @@ outputs:
...
-        {% set skips = skips ~ " or test_conv_nhwc_lower_precision_cpu_float16" %}       # [aarch64]
-        {% set skips = skips ~ " or test_conv_transpose_nhwc_lower_precision_cpu_float16" %}  # [aarch64]
-        {% set skips = skips ~ " or test_matmul_lower_precision_cpu_float16" %}          # [aarch64]
-        {% set skips = skips ~ " or (test_save_load_nn_ConvTranspose and cpu_complex32)" %}  # [aarch64]
-        {% set skips = skips ~ " or (test_if_train_and_eval_modes_differ_nn_ConvTranspose and cpu_complex32)" %}  # [aarch64]
-        {% set skips = skips ~ " or test_memory_format_nn_ConvTranspose1d_cpu_complex32" %}  # [aarch64]
-        {% set skips = skips ~ " or (test_non_contiguous_tensors_nn_ConvTranspose and cpu_complex32)" %}  # [aarch64]
-        {% set skips = skips ~ " or test_activations_bfloat16_half_cpu_cpu_float16" %}   # [aarch64]
-        {% set skips = skips ~ " or test_linear_lowp_cpu_float16" %}                     # [aarch64]
-        {% set skips = skips ~ " or test_eigh_svd_illcondition_matrix_input_should_not_crash_cpu_float32" %}  # [aarch64]
```
mgorny, line 568: Looks like this is still necessary.

recipe/meta.yaml
```diff
@@ -548,41 +575,19 @@ outputs:
...
```
mgorny, line 579:
```suggestion
        {% set skips = skips ~ " or test_eigh_svd_illcondition_matrix_input_should_not_crash_cpu_float32" %}  # [aarch64]

```

PR #509, opened 2026-06-07
V2.12 (hmaarrfk)

hmaarrfk:
I used claude (and reviewed much of its work) to try to get the 2.12 branch up and runnig.

I think this is in good shape.

I removed the addition of the `CLAUDE.md` file that helped me overcome the challenges.

Checklist
* [ ] Used a [personal fork of the feedstock to propose changes](https://conda-forge.org/docs/maintainer/updating_pkgs.html#forking-and-pull-requests)
* [ ] Bumped the build number (if the version is unchanged)
* [ ] Reset the build number to `0` (if the version changed)
* [ ] [Re-rendered]( https://conda-forge.org/docs/maintainer/updating_pkgs.html#rerendering-feedstocks ) with the latest `conda-smithy` (Use the phrase <code>@<space/>conda-forge-admin, please rerender</code> in a comment in this PR for automated rerendering)
* [ ] Ensured the license file is being packaged.

PR #512, opened 2026-06-11
PyTorch v2.12.0

wolfv:
Builds on the work in #509 (this branch includes all of its commits — thanks @hmaarrfk, @mgorny, @h-vetinari and everyone involved!) and adds the remaining fixes to get CI green:

### Fixes on top of #509

- **Skip `TestNN::test_module_backcompat`** — the only failing test in the otherwise-green CI run on #509 (win/CUDA 13.0). It downloads a legacy serialized module from `download.pytorch.org` and failed with an `ssl.SSLError`, i.e. a network flake that can hit any platform.
- **Patch 0022: add missing `<unordered_map>`/`<vector>` includes** in `torch/csrc/distributed/c10d/control_plane/Handlers.hpp` (fails to compile on osx-arm64 metal builds; from AnacondaRecipes).
- **Patch 0023: remove the deprecated `path` parameter** from `pytest_pycollect_makemodule` in `test/conftest.py` (removed in pytest 9; from AnacondaRecipes), and **drop the `pytest <9` pin** that worked around it.

### Notes from comparing with AnacondaRecipes' 2.12 work

Their `2.12-prep` branch was reviewed for anything else we might need; the rest doesn't apply to us:
- tensorpipe CUDA 13 `computeMode` fix — already fixed in the submodule commit pinned by v2.12.0
- oneDNN patches — they force the bundled oneDNN, we use the system one
- pybind11 3.x / `pybind11-abi 11` — already handled by our pinning
- fbgemm/CUTLASS job pools, nvcc memory flags — resource tuning for their runners; our CUDA builds pass without it

🤖 Generated with [Claude Code](https://claude.com/claude-code)


recipe/meta.yaml
```diff
@@ -483,6 +499,12 @@ outputs:
         {% set skips = skips ~ " or test_grad_nn_Transformer" %}                        # [aarch64]
         {% set skips = skips ~ " or test_inverse_errors_large" %}                       # [aarch64]
         {% set skips = skips ~ " or (TestXNNPACKConv1dTransformPass and test_conv1d_basic)" %}  # [aarch64]
+        # errors (possibly QEMU-related) with openblas 0.3.30
+        {% set skips = skips ~ " or test_addbmm or test_baddbmm or test_bmm" %}         # [aarch64]
```
h-vetinari: This looks like a merge left-over; these skips are now in
https://github.com/conda-forge/pytorch-cpu-feedstock/blob/5f37bfc40597d26ceb3ed8a059b0691bb06b9b9d/recipe/meta.yaml#L452-L458

recipe/patches/0024-Use-serializable-subtest-labels-in-autograd.patch
```diff
@@ -0,0 +1,33 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: conda-forge <conda-forge@users.noreply.github.com>
```
h-vetinari: If you're taking this from Anaconda's recipes, can you keep the attribution (which, at the very least, can be gleaned from whichever commit originally introduced any given patch)?

recipe/patches/0022-Add-missing-includes-in-c10d-control_plane-Handlers.patch
```diff
@@ -0,0 +1,27 @@
+From ac569bd0ba27d222e26fc2dd5671e47e4dcbd7cc Mon Sep 17 00:00:00 2001
+From: conda-forge <conda-forge@users.noreply.github.com>
```
h-vetinari: same

recipe/patches/0019-Fix-dispatch-key-guard-leak-in-test_incorrect_abstr.patch
```diff
@@ -0,0 +1,60 @@
+From 0aba567e2e89cf0da6918f3b8fb3d8f3f87e0cda Mon Sep 17 00:00:00 2001
+From: Mark Harfouche <mark@ramonaoptics.com>
+Date: Sun, 7 Jun 2026 15:21:44 +0000
+Subject: [PATCH] Fix dispatch-key guard leak in test_incorrect_abstract_impl
```
h-vetinari: @hmaarrfk, this patch sounds like it should be upstreamed? Would you like to open a PR to pytorch?

recipe/patches/0002-Revert-Move-environment-variable-handling-from-setup.patch
```diff
@@ -0,0 +1,368 @@
+From e5f013c16bc1bee02138a9414194a42c42885d23 Mon Sep 17 00:00:00 2001
+From: =?UTF-8?q?Micha=C5=82=20G=C3=B3rny?= <mgorny@quansight.com>
+Date: Thu, 4 Jun 2026 18:31:24 +0200
+Subject: [PATCH 02/17] Revert "Move environment variable handling from
+ setup.py to CMake (#177640)"
```
h-vetinari: This does not sound like a durable solution, as that revert will become harder-and-harder to apply, much less to keep it correct. I'd be surprised if upstream stopped caring about cross-compilation at all... Could you raise an issue upstream @mgorny?

h-vetinari (approved):
Thanks for everyone's efforts on this! 🙏 

This LGTM up to some minor nits. Please provide the correct patch attributions in a comment on the respective thread, I can fix things up while merging.

recipe/patches/0002-Revert-Move-environment-variable-handling-from-setup.patch
mgorny: Yeah, it's only temporary. Already highlighted the issue to Klaus.

recipe/patches/0019-Fix-dispatch-key-guard-leak-in-test_incorrect_abstr.patch
hmaarrfk: I do not have time to understand the details of this change in particular to upstream it.

h-vetinari: So the diagnosis & commit message is from an LLM, presumably?

hmaarrfk: I just don’t have time to drill down to the correct place to put it. 

I haven’t shied away from the fact that an LLM made 100% of the commits. 

But I stand behind “escaping a few slashes to get through the build process”

hmaarrfk: PyTorch upstream has also been quite non receptive to commits that don’t benefit their build system.

recipe/meta.yaml
wolfv: Good catch, thanks — that line was indeed accidentally resurrected in a merge (it was removed in 6745144 when CI moved to native agents). Dropped in b895644.

recipe/patches/0022-Add-missing-includes-in-c10d-control_plane-Handlers.patch
wolfv: Done in b895644 — restored the original authorship (Daniel Petry, from AnacondaRecipes/pytorch-feedstock) and noted the provenance in the patch description; same note added to 0023, which already carried its original author.

recipe/patches/0024-Use-serializable-subtest-labels-in-autograd.patch
wolfv: This one isn't from Anaconda's recipes — it was written for this PR (to fix pytest-xdist failing to serialize subtest reports that had torch builtins as parameters). The `conda-forge` author was a placeholder; b895644 sets the actual authorship. Happy to upstream it to pytorch, it should be uncontroversial.

h-vetinari:
Hm, I thought the errors à la
```
  File "/Users/runner/miniforge3/lib/python3.12/site-packages/conda_build/render.py", line 496, in _read_upstream_pin_files
    extra_run_specs = get_upstream_pins(m, precs, env)
                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/runner/miniforge3/lib/python3.12/site-packages/conda_build/render.py", line 466, in get_upstream_pins
    loc, dist = execute_download_actions(
                ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/runner/miniforge3/lib/python3.12/site-packages/conda_build/render.py", line 421, in execute_download_actions
    link_prec = [
                ^
IndexError: list index out of range
```
with your last push were transient (so I took the opportunity to regenerate the patches), but alas that doesn't seem to be the case. Though usually this kind of error has always disappeared upon restarts shortly after.

(yes, yes, I know, it's time for the v1 conversion 😋 - PRs welcome!)

wolfv:
@h-vetinari looks like it's still not quite correct - some patch is missing. Do you want me to take a look or will you push another commit?

h-vetinari:
Thanks for the great teamwork on this 💪

PR #517, opened 2026-06-12
[DEBUG] 2.12.0 with EnvVarForwarding

zklaus:
Another 2.12 variation :duck:. This one keeps the EnvVarForwarding intact, instead injecting CMAKE_ARGS in the build process, which aligns more closely with how scikit-build-core will be handling this.


wolfv:
If possible maybe not spin up windows jobs as we are reaching the $500 limit we set recently :/ :D

h-vetinari:
> If possible maybe not spin up windows jobs as we are reaching the $500 limit we set recently :/ :D

We should definitely move the windows runners away from cirun and towards one of the [providers](https://github.com/conda-forge/admin-requests#larger-runners-for-github-actions) for larger runners

zklaus:
With v2.12 merged, the rest of this PR is now in #520.

PR #518, opened 2026-06-12
Port to v1 (rattler-build) recipe with staging outputs + Python test runner

wolfv:
Builds on the 2.12 work in #512, but converts the recipe to the **v1 (rattler-build)** format to get out of the brittle conda-build multi-output finalize that #512 keeps tripping over (the latest failure there is a conda-build \`IndexError\` in \`get_upstream_pins\`, not a recipe-solve error).

## What changed

- **\`recipe/meta.yaml\` → \`recipe/recipe.yaml\`** (v1).
- **Staging output** (\`libtorch-build\`) compiles the heavy C++ **once** per (target_platform, cuda, blas); the per-python \`pytorch\` output \`inherit\`s the staging build tree and only rebuilds the python bindings on top of it. This preserves the old "build libtorch once, reuse for each python" optimization that the \`PKG_NAME\`-branching \`build.sh\` used to get via shared work dirs — render confirms all per-python \`pytorch\`/\`pytorch-tests\` builds pin the *same* libtorch hash.
- **\`recipe/run_pytorch_tests.py\`** replaces the ~180-line \`{% set skips %}\` Jinja swamp. Test files + skips are plain Python data; each platform selector is an ordinary boolean. Variant is passed via env vars (\`PYTORCH_VARIANT_*\`) with runtime fallback.
- **\`build.sh\`/\`build.bat\`** keep their two stages, now selected via \`PYTORCH_BUILD_STAGE\` (set by the recipe) instead of \`PKG_NAME\`. \`bld.bat\` renamed to \`build.bat\` per rattler-build convention.
- **\`conda-forge.yml\`**: \`conda_build_tool: rattler-build\`, \`conda_install_tool: pixi\`.

Outputs preserved: \`libtorch\`, \`pytorch\`, \`pytorch-tests\`, \`pytorch-cpu\`/\`pytorch-gpu\`. Build-number offsets (cuda +200, mkl +100) and the rc-skip logic are preserved.

Renders cleanly across linux-64/aarch64 (cpu+cuda), osx-arm64, win-64; \`conda-smithy lint\` + rerender pass. A local osx-arm64 cpu build is running to exercise the staging split + test runner end-to-end.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


recipe/run_pytorch_tests.py
h-vetinari, line 1: > replaces the ~180-line `{% set skips %}` Jinja swamp

Nope. Test skips belong in the recipe, not in a sidecar. You can use the [numpy approach](https://github.com/conda-forge/numpy-feedstock/blob/c4afd1ee0220c76e360a4c3c240d0db850fc8724/recipe/recipe.yaml#L137-L160)

h-vetinari (changes requested):

.ci_support/linux_aarch64_channel_targetsconda-forge_maincuda_compiler_version12.9is_rcFalse.yaml
```diff
@@ -1,14 +1,17 @@
 arm_variant_type:
+- sbsa
 - tegra
```
h-vetinari, line 3: This is incorrect; needs to be two separate jobs

recipe/recipe.yaml
```diff
@@ -0,0 +1,688 @@
...
+  # that conda's solver prefers them where the platform supports them (this
+  # mirrors the old meta.yaml build-number juggling).
+  build_base: 0
+  build_number: ${{ build_base + (200 if cuda_compiler_version != "None" else 0) + (100 if blas_impl == "mkl" else 0) }}
+  # see https://github.com/pytorch/pytorch/blame/v${{ version }}/.ci/docker/ci_commit_pins/triton.txt
+  # pytorch and triton are released in tandem, see notes in their release process
+  # https://github.com/pytorch/pytorch/blob/main/RELEASE.md#triton-dependency-for-the-release
+  triton: "3.7.0"
+  release_sha256: "7cc1deb309f402ad67e9f45bbe311a40def4db19d66fddb9b01950f9bfc5ccb1"
+  rc_sha256: "3035931fff5b79e0300db69b0249e0c9c7ea5b394f451cb80e0280cbc2affcc2"
```
h-vetinari: I'd prefer to just have a single sha, directly under `source.sha256`; when folks switch between rc & non-rc, they just need to update the sha.

recipe/recipe.yaml
```diff
@@ -0,0 +1,688 @@
...
+            - libcusolver-dev
+            - libcusparse-dev
+            - libmagma-devel
+            - nvtx-c
+        - if: linux and cuda_compiler_version != "None"
+          then:
+            - cuda-driver-dev
+            - libcufile-dev
+            - nccl
+            - if: cuda_compiler_version == "12.9" and (arm_variant_type | default("sbsa")) != "tegra"
```
h-vetinari: prefer positive comparisons to inverted ones
```suggestion
            - if: cuda_compiler_version == "12.9" and (arm_variant_type | default("sbsa")) == "sbsa"
```

recipe/recipe.yaml
```diff
@@ -0,0 +1,688 @@
...
+        missing_dso_allowlist:
+          - "*/libcuda.so.*"
+          - "*/nvcuda.dll"
+      prefix_detection:
+        ignore_binary_files: true
+      files:
+        # everything the staging C++ build produced, i.e. the non-python bits
+        - if: unix
+          then:
+            - lib/lib*
```
h-vetinari: Don't you have to exclude `lib/libtorch_python*` (and so on for windows) here for this to work?

recipe/recipe.yaml
```diff
@@ -0,0 +1,688 @@
...
+        # only the python-specific bits; the C++ libs belong to libtorch
+        - if: unix
+          then:
+            - lib/python*/site-packages/torch/**
+            - lib/libtorch_python*
+          else:
+            - Lib/site-packages/torch/**
+            - Library/bin/torch_python.dll
+            - Library/lib/torch_python.lib
+            - Library/lib/_C.lib
```
h-vetinari: The entire `files:` spec here is redundant/contradictory with the fact that we have a build script (+ `PYTORCH_BUILD_STAGE`) that already installs things. In fact, I'd prefer for the `libtorch` output to also call the build-scripts again (with a suitable `PYTORCH_BUILD_STAGE` that just continues the install from the staging output).

conda-forge.yml
```diff
@@ -4,13 +4,15 @@ azure:
...
+  - v2.9.x
 build_platform:
   linux_aarch64: linux_64
   osx_64: osx_arm64
 conda_build:
   pkg_format: '2'
   zstd_compression_level: 22
+conda_build_tool: rattler-build
+conda_install_tool: pixi
 conda_forge_output_validation: true
```
h-vetinari, line 15: please keep this alphabetical
```suggestion
conda_build_tool: rattler-build
conda_forge_output_validation: true
conda_install_tool: pixi
```

recipe/recipe.yaml
```diff
@@ -0,0 +1,688 @@
...
+          - numpy
+          - libmagma_sparse
+
+  # ----------------------------------------------------------------------- #
+  # Core C++ library. Just repackages the C++ artifacts from the staging build.
+  # ----------------------------------------------------------------------- #
+  - package:
+      name: libtorch
+    build:
+      string: ${{ flavor }}_${{ hash }}_${{ build_number }}
```
h-vetinari, line 241: Where does `${{ hash }}` come from? Is this a magic variable? 🤔

recipe/recipe.yaml
wolfv: Done

recipe/recipe.yaml
wolfv: Done

recipe/recipe.yaml
wolfv: Restructured into 3 stages: staging build compiles once; libtorch and pytorch outputs run the build script (PYTORCH_BUILD_STAGE=libtorch/pytorch) to install their files. Removed all files: globs. Updated build.sh and build.bat.

conda-forge.yml
wolfv: Done!

recipe/recipe.yaml
wolfv: Yes, that is a rattler-build variable indeed. (replaces `PKG_HASH`)

h-vetinari:
Sorry, I cancelled CI here (there's already a bunch of failures which should provide some insights), in order to have maximum allocated workers on the first go for merging #512

recipe/recipe.yaml
h-vetinari: What happens if someone set `hash: "foo"` in the `context:`? 🙃 

It's got a pretty innocuous (I'd say _too_ innocuous) name for my taste 😅

recipe/recipe.yaml
```diff
@@ -0,0 +1,841 @@
...
+  # see https://github.com/pytorch/pytorch/blame/v${{ version }}/.ci/docker/ci_commit_pins/triton.txt
+  # pytorch and triton are released in tandem, see notes in their release process
+  # https://github.com/pytorch/pytorch/blob/main/RELEASE.md#triton-dependency-for-the-release
+  triton: "3.7.0"
+  # build string fragment: cpu vs cuda + blas implementation
+  flavor: ${{ "cuda" ~ (cuda_compiler_version | replace(".", "")) if cuda_compiler_version != "None" else "cpu" }}_${{ blas_impl }}
+
+recipe:
+  # name is ignored for multi-output recipes
+  name: pytorch-cpu
```
mgorny, line 20:
```suggestion
  name: pytorch-split
```

I think that's the common convention.

recipe/recipe.yaml
```diff
@@ -0,0 +1,841 @@
...
+  # normalize possible rc-style versions the same way the old recipe did
+  version: ${{ version | replace("-", ".") }}
+
+source:
+  # The "rc" tarballs live on download.pytorch.org; releases are GitHub assets.
+  # The "pytorch-v" tarballs contain submodules; the "pytorch-" ones don't.
+  - url: ${{ "https://download.pytorch.org/source_code/test/pytorch-v" ~ version ~ ".tar.gz" if "rc" in version else "https://github.com/pytorch/pytorch/releases/download/v" ~ version ~ "/pytorch-v" ~ version ~ ".tar.gz" }}
+    # single sha; update this when switching between rc and final releases
+    sha256: 7cc1deb309f402ad67e9f45bbe311a40def4db19d66fddb9b01950f9bfc5ccb1
+    patches:
```
mgorny, line 30: Just leaving a note that this at least needs updating after rebase.

recipe/recipe.yaml
```diff
@@ -0,0 +1,841 @@
...
+            - python 3.12.*
+            - cross-python_${{ target_platform }}
+            - numpy *
+        - ${{ stdlib("c") }}
+        - ${{ compiler("c") }}
+        - ${{ compiler("cxx") }}
+        - if: cuda_compiler_version != "None"
+          then:
+            - ${{ compiler("cuda") }}
+        - if: linux and aarch64 and cuda_compiler_version != "None"
```
mgorny, line 131: Nest that inside `cuda_compiler_version != "None"`.

recipe/recipe.yaml
```diff
@@ -0,0 +1,841 @@
...
+            - libcublas-dev
+            - libcudnn-dev
+            - libcudss-dev
+            - libcufft-dev
+            - libcurand-dev
+            - libcusolver-dev
+            - libcusparse-dev
+            - libmagma-devel
+            - nvtx-c
+        - if: linux and cuda_compiler_version != "None"
```
mgorny, line 175: Likewise.

recipe/recipe.yaml
```diff
@@ -0,0 +1,841 @@
...
+            - shm
+            - torch
+            - torch_cpu
+            - torch_global_deps
+            - if: cuda_compiler_version != "None"
+              then:
+                - c10_cuda
+                - caffe2_nvrtc
+                - torch_cuda
+            - if: cuda_compiler_version != "None" and linux
```
mgorny, line 309: Nest.

recipe/recipe.yaml
```diff
@@ -0,0 +1,841 @@
...
+        files:
+          recipe:
+            - cmake_test
+        script:
+          # test integrity of CMake metadata
+          - cd cmake_test
+          - if: unix
+            then: cmake -GNinja -DCMAKE_CXX_STANDARD=17 $CMAKE_ARGS .
+          - if: win
+            then: cmake -GNinja -DCMAKE_CXX_STANDARD=17 %CMAKE_ARGS% .
```
mgorny, line 341:
```suggestion
            else: cmake -GNinja -DCMAKE_CXX_STANDARD=17 %CMAKE_ARGS% .
```

recipe/recipe.yaml
```diff
@@ -0,0 +1,841 @@
...
+            - ninja
+            - pip
+            - numpy
+        files:
+          recipe:
+            - cmake_test
+        script:
+          # Ensure that all pytorch packages are installed
+          # https://github.com/conda-forge/pytorch-cpu-feedstock/issues/24
+          - pip check
```
mgorny, line 724: Can't we do the `python:` testing here?

mgorny (commented):
To be honest, this is a really bad time to be doing this, given how much churn's on main.

mgorny:
@wolfv, before you start working on this again, please rebase on `main` so that you use the new Namespace runners and not CIrun.

recipe/recipe.yaml
h-vetinari: It's a convention that's (much) less relevant for v1 recipes; v0 recipes do strange things when the top-level package shares the same name as an `output:`, and so adding `-split` to the former emerged as a quasi-standard organically; v1 improves on this, so the old customs serve no purpose here anymore.

recipe/recipe.yaml
h-vetinari: not sure what the exact rules are, but I tend to (single-)quote lines containing `%` characters just to be on the safe side.

recipe/recipe.yaml
```diff
@@ -0,0 +1,841 @@
...
+  - package:
+      name: pytorch-gpu
+    build:
+      string: ${{ flavor }}_${{ hash }}_${{ build_number }}
+      skip: ${{ cuda_compiler_version == "None" }}
+    requirements:
+      host:
+        - if: linux and aarch64 and cuda_compiler_version != "None"
+          then:
+            - arm-variant * ${{ arm_variant_type }}
```
h-vetinari, line 805: For all outputs other than `libtorch` (and maybe `pytorch`), the `arm-variant` host-dep was only there to ensure jobs got split along this axis. We should use something like a top-level
```yaml
build:
  variant:
    use_keys:
      - arm_variant_type
```
here and then remove this host dep from tests and wrapper outputs

recipe/recipe.yaml
```diff
@@ -0,0 +1,841 @@
...
+      run:
+        - pytorch ==${{ version }} cuda*_${{ blas_impl }}*${{ build_number }}
```
h-vetinari, line 807: If we want to pin this tightly, we might as well reuse `${{ flavor }}` here (since it's already part of the `build.string:`). My preferred approach for the convenience wrappers though would be to not distinguish them between blas variant and/or CUDA version. The very existence of `pytorch-gpu` as a convenience is to avoid doing build-string-level pinning, so then adding that to the wrapper kinda defeats the purpose. I'd rather do something like
```suggestion
      run:
        # independent of blas implementation or CUDA version
        - pytorch ==${{ version }} cuda*_${{ build_number }}
```

recipe/recipe.yaml
h-vetinari: Hm. I kinda prefer not nesting (not to the point of arguing about it, but just to say that there's an argument to be made to avoid even deeper nesting levels)

recipe/recipe.yaml
h-vetinari: See my other comment. I'd keep the `if:`s on a single level, but order them consistently by their conditions, e.g.
```suggestion
        - if: cuda_compiler_version != "None" and linux and aarch64
```

recipe/recipe.yaml
```diff
@@ -0,0 +1,841 @@
...
+        - if: linux and cuda_compiler_version != "None"
+          then:
+            - cuda-driver-dev
+            - libcufile-dev
+            - nccl
+            - if: cuda_compiler_version == "12.9" and (arm_variant_type | default("sbsa")) == "sbsa"
+              then:
+                - nccl <2.29
+        # other requirements
+        - python 3.12.*
```
h-vetinari, line 184: It'd be nice to move the 3.12 to the `context:`, and then refactor
https://github.com/conda-forge/pytorch-cpu-feedstock/blob/709fc4c6a01112a466d270336cdf338ab99a964c/recipe/build.sh#L169-L172
to re-use that.

h-vetinari (commented):
Things have settled down a bit, so this could use a rebase. :)

recipe/recipe.yaml
h-vetinari: > The entire `files:` spec here is redundant/contradictory with the fact that we have a build script

Let me clarify that a bit better; using general build-scripts (i.e. ones that do variations of `make install` for every output) together with `files:` can still make sense. But once the build scripts take care of the separation, it's _mostly_ redundant to list it also in the recipe. However, I've come around to seeing the benefit of being explicit about `files:` in the recipe, as it provides a line-of-defense against errors in the build scripts.

PR #523, opened 2026-06-14
[main] Rebuild for libabseil 20260526 & libprotobuf 7.35.1 

h-vetinari:
Preparing in advance for the next abseil/proto migration. Also picking up #520, which is a super-nice clean-up, and #519, which accidentally showed that the errors with GCC 14 on aarch are now gone after v2.11.

Closes #520

PR #524, opened 2026-06-17
Switch to Namespace runners for windows

mgorny:


mgorny:
(I've added `free_disk_space` only to get a `df` in.)

h-vetinari:
This looks pretty good, aside from running into [issues](https://conda.zulipchat.com/#narrow/channel/487124-infrastructure/topic/too.20many.20requests.20for.20repodata.3F/with/604280023) with Anaconda's infra! I also need to push a fix for newest pytest, because the last builds from #523 never made it. Might as well try it.

> (I've added `free_disk_space` only to get a `df` in.)

Looks pretty generous! (but crazy how there's already 180GB in `C:`... I guess a lot of this would be clean-up-able)
```
+ df -h
Filesystem            Size  Used Avail Use% Mounted on
C:/Program Files/Git  200G  178G   23G  89% /
D:                    1.2T  199M  1.2T   1% /d
```

mgorny:
> (but crazy how there's already 180GB in C:... I guess a lot of this would be clean-up-able)

Yeah, probably — but we use D:, so…

PR #525, opened 2026-06-18
pytorch v2.12.1

h-vetinari:
New patch release; CUDA will have to wait for builds from https://github.com/conda-forge/triton-feedstock/pull/72

### DIFF
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -1,6 +1,6 @@
 # if you wish to build release candidate number X, append the version string with "-rcX"
-{% set version = "2.12.0" %}
-{% set build = 1 %}
+{% set version = "2.12.1" %}
+{% set build = 0 %}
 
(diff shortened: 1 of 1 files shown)


hmaarrfk:
Nice. Thanks!

mgorny:
Cool that Namespace runners worked! Now we just need to figure out GPU testing and the Infra should be real good :-).

mgorny:
Are the triton errors a race between these jobs and the uploads, or am I missing a deeper issue?

h-vetinari:
> Are the triton errors a race between these jobs and the uploads, or am I missing a deeper issue?

Yup, nothing more than that. TBH I expected the CUDA jobs to fail right away, but it appears `triton` comes in as a dependency only later, so the builds continued, and some even finished, because by then `triton` builds where available. I've restarted the remaining failures

PR #526, opened 2026-07-01
2.13.0

mgorny:
Checklist


mgorny:
Windows seems to be failing over:

```
  -- Unable to find cublas_v2.h in either "D:/bld/libtorch_1782915765464/_build_env/Library/include" or "D:/bld/libtorch_1782915765464/math_libs/include"
```

Looks like it's using build paths instead of host paths.

mgorny:
linux-aarch64 jobs are failing with `ImportError`. I'm going to add a patch to reraise the original `ImportError` to investigate.

osx-64 jobs are failing over:

```
2026-07-01T14:40:19.8643980Z Undefined symbols for architecture x86_64:
2026-07-01T14:40:19.8644310Z   "at::native::DispatchStubImpl::get_call_ptr(c10::DeviceType, void*)", referenced from:
2026-07-01T14:40:19.8644840Z       at::native::structured_gelu_out_mps::impl(at::Tensor const&, std::__1::basic_string_view<char, std::__1::char_traits<char>>, at::Tensor const&) in Activation.mm.o
2026-07-01T14:40:19.8645490Z       at::native::structured_gelu_backward_out_mps::impl(at::Tensor const&, at::Tensor const&, std::__1::basic_string_view<char, std::__1::char_traits<char>>, at::Tensor const&) in Activation.mm.o
2026-07-01T14:40:19.8646030Z      (maybe you meant: at::native::DispatchStubImpl::get_call_ptr(c10::DeviceType, void*, void*, void*))
2026-07-01T14:40:19.8646310Z ld: symbol(s) not found for architecture x86_64
```

I see the arguments there depend on a bunch of defines such as `HAVE_AVX512_CPU_DEFINITION` and `HAVE_AVX2_CPU_DEFINITION`; I'm guessing these aren't getting consistently defined when cross-compiling. Need to investigate.

mgorny:
Uh, looks like the C extension isn't built at all on AArch64 for some reason…

mgorny:
Or it is built, but it's broken. And doesn't get the usual suffix.

mgorny:
Okay, I've just noticed that the logic for linking the C extension was moved from setuptools to CMake in this version, so I'm wrong to be comparing the failed AArch64 builds to earlier artifacts. I need to check how "working" linux-64 artifacts look now.

isuruf:
> Okay, I've just noticed that the logic for linking the C extension was moved from setuptools to CMake in this version, 

That's @zklaus's work I think. ccing him here if you need help.

mgorny:
Already gave him a heads up on Slack, but I'm still debugging. So far my conclusion is that CMake's logic can't figure out `SOABI` when cross-compiling. I'm checking right now if my earlier override of `Python_ROOT_DIR` may have been causing this.

mgorny:
I can reproduce the error via wine, but I'm not sure if it's actually the same problem or incidental problem with wine:

```
0228:err:module:import_dll Library cupti64_2025.3.1.dll (which is needed by L"z:\\srv\\p2p\\conda\\miniforge3\\envs\\pytorch-test\\Library\\bin\\torch_cpu.dll") not found
0228:err:module:import_dll Library torch_cpu.dll (which is needed by L"z:\\srv\\p2p\\conda\\miniforge3\\envs\\pytorch-test\\Library\\bin\\torch_cuda.dll") not found
0228:err:module:import_dll Library torch_cuda.dll (which is needed by L"z:\\srv\\p2p\\conda\\miniforge3\\envs\\pytorch-test\\Library\\bin\\shm.dll") not found
0228:err:module:import_dll Library cupti64_2025.3.1.dll (which is needed by L"z:\\srv\\p2p\\conda\\miniforge3\\envs\\pytorch-test\\Library\\bin\\torch_cpu.dll") not found
0228:err:module:import_dll Library torch_cpu.dll (which is needed by L"z:\\srv\\p2p\\conda\\miniforge3\\envs\\pytorch-test\\Library\\bin\\shm.dll") not found
0228:err:module:import_dll Library shm.dll (which is needed by L"z:\\srv\\p2p\\conda\\miniforge3\\envs\\pytorch-test\\Library\\bin\\torch_python.dll") not found
0228:err:module:import_dll Library cupti64_2025.3.1.dll (which is needed by L"z:\\srv\\p2p\\conda\\miniforge3\\envs\\pytorch-test\\Library\\bin\\torch_cpu.dll") not found
0228:err:module:import_dll Library torch_cpu.dll (which is needed by L"z:\\srv\\p2p\\conda\\miniforge3\\envs\\pytorch-test\\Library\\bin\\torch_cuda.dll") not found
0228:err:module:import_dll Library torch_cuda.dll (which is needed by L"z:\\srv\\p2p\\conda\\miniforge3\\envs\\pytorch-test\\Library\\bin\\torch_python.dll") not found
0228:err:module:import_dll Library cupti64_2025.3.1.dll (which is needed by L"z:\\srv\\p2p\\conda\\miniforge3\\envs\\pytorch-test\\Library\\bin\\torch_cpu.dll") not found
0228:err:module:import_dll Library torch_cpu.dll (which is needed by L"z:\\srv\\p2p\\conda\\miniforge3\\envs\\pytorch-test\\Library\\bin\\torch_python.dll") not found
0228:err:module:import_dll Library torch_python.dll (which is needed by L"z:\\srv\\p2p\\conda\\miniforge3\\envs\\pytorch-test\\Lib\\site-packages\\torch\\_C.cp314-win_amd64.pyd") not found
  File "<python-input-0>", line 1, in <module>
    import torch
  File "z:\srv\p2p\conda\miniforge3\envs\pytorch-test\Lib\site-packages\torch\__init__.py", line 445, in <module
>
    from torch._C import *  # noqa: F403
    ^^^^^^^^^^^^^^^^^^^^^^
ImportError: DLL load failed while importing _C: Module not found.
```

If it's the same, that would mean that somehow the DLL paths aren't set correctly for the Python extension. That said, I have no clue how this works on Windows, but given that the extension build process changed, it might indeed be the case.

Could someone more familiar with Windows help?

isuruf:
Opened https://github.com/conda-forge/cuda-cupti-feedstock/issues/39

recipe/meta.yaml
```diff
@@ -352,6 +352,9 @@ outputs:
         - setuptools <82
         - sympy >=1.13.3
         - typing_extensions >=4.10.0
+        # cuda-cupti-dev used to be pinned too widely on WIndows
+        - cuda-cupti 12.8.*                 # [win and cuda_compiler_version=="12.8"]
+        - cuda-cupti 13.0.*                 # [win and cuda_compiler_version=="13.0"]
```
isuruf: You can do
```suggestion
        - {{ pin_compatible("cuda-cupti", max_pin="x.x.x") }}          # [win and cuda_compiler_version != "None"]
```

mgorny: Ok, I'll integrate this and rebase.

mgorny:
Updated the define patch to the new upstream version, skipped the failing test and now retrying the full set again.

mgorny:
Sigh, what's that regression now?

mgorny:
Looks like it's not finding the right Python now.

Before:

```
-- Found Python: $PREFIX/bin/python3.10 (found version "3.10.20") found components: Interpreter
```

Now:

```
-- Found Python: /usr/bin/python3.12 (found version "3.12.13") found components: Interpreter
```

Though at least one another impl passed before that, and I don't see any relevant changes in dependencies.

mgorny:
Uh, had the checksum updated locally but forgot to commit it :facepalm:.

mgorny:
Okay, so the CMake failure is because CMake now defaults to trying to use the newest Python version it can find rather than one from the top location; so if there is a system Python that's newer than the tested Python, it takes that.

We can work around the problem by setting:

```
set(Python3_FIND_STRATEGY LOCATION)
```

…but I do wonder if that's not a symptom of a bigger problem. I mean, why the sudden change? It also happens when I `find_package(Python3 ...)` before anything else.

isuruf:
I guess the change from Alma9 to Alma10 changed the system python version and is now found.

mgorny:
Oh, yes, that would explain it. Which opens a subsequent question: should we be patching CMake to prefer conda-forge Python over system Python? Should I open a bug on cmake-feedstock about this?

isuruf:
Maybe we should add `Python3_FIND_STRATEGY=LOCATION` in `CMAKE_ARGS` set by the compiler. I'm not a fan of patching CMake when the behaviour is clearly defined in the docs.

isuruf:
See https://github.com/conda-forge/cmake-feedstock/issues/196

mgorny:
> Maybe we should add `Python3_FIND_STRATEGY=LOCATION` in `CMAKE_ARGS` set by the compiler. I'm not a fan of patching CMake when the behaviour is clearly defined in the docs.

Yeah, I'll do that when the running jobs finish, as they don't have much time left, and I confirm that there are no other regressions.

Alternatively, we could also explicitly specify the Python version, like we do in `python-feedstock`.

isuruf:
> Maybe we should add Python3_FIND_STRATEGY=LOCATION in CMAKE_ARGS set by the compiler.

I meant in the activation scripts of the compiler

recipe/cmake_test/CMakeLists.txt
```diff
@@ -1,6 +1,9 @@
 project(cf_dummy LANGUAGES C CXX)
 cmake_minimum_required(VERSION 3.12)
```
isuruf: You can also use this to enable CMP0094
```suggestion
cmake_minimum_required(VERSION 3.15)
```

mgorny: Good idea either way, thanks!

mgorny:
It is green!

h-vetinari:
Thanks for this @mgorny! 🙏 

Could you open a migrator for 2.13? We should also try to wind down https://conda-forge.org/status/migration/?name=pytorch212

mgorny:
Sure, I'll open it. I'll also try to find some time to look at the older migrations, but that's after higher priority stuff.

mgorny:
https://github.com/conda-forge/conda-forge-pinning-feedstock/pull/8780

Issue #447, opened 2025-11-10
patch for include paths wrongly points to BUILD_PREFIX in cross-compilation


mikemhenry:
This patch: https://github.com/conda-forge/pytorch-cpu-feedstock/blob/56a9a5b255b0037985d9f077fc88459661a0792b/recipe/patches/0008-point-include-paths-to-Conda-prefix-include-dir.patch#L27

Adds `$BUILD_PREFIX/include` to the `-I` flag which causes lots of errors like `error: redefinition of 'OrderedDict'` down stream, see https://github.com/conda-forge/pytorch_scatter-feedstock/pull/74

hmaarrfk:
it is suspect to me that the build_prefix is found in the include path at all.

h-vetinari:
@mikehenry, did you ever try this out? I just tried https://github.com/conda-forge/torchvision-feedstock/pull/138 and we still end up picking up the includes from the `$BUILD_PREFIX` 😑 

It's not from `$BUILD_PREFIX/include` though, but actually from the site-packages directory...

```
 │ │   $BUILD_PREFIX/venv/lib/python3.11/site-packages/torch/include/torch/csrc/api/include/torch/types.h:69:16: error: conflicting declaration 'constexpr const auto torch::kF64'
 │ │      69 | constexpr auto kF64 = kFloat64;
 │ │         |                ^~~~
 │ │   $PREFIX/include/torch/csrc/api/include/torch/types.h:69:16: note: previous declaration as 'constexpr const c10::ScalarType torch::kF64'
 │ │      69 | constexpr auto kF64 = kFloat64;
 │ │         |                ^~~~
```

I'd be tempted to just delete those headers in `pytorch` (rather than in `libtorch`), but presumably some builds might rely on that location (and adding a symlink to `$PREFIX/include` - while unequivocally better on unix - would still end up getting found)

h-vetinari:
Also seeing this elsewhere, e.g. https://github.com/conda-forge/torch-scatter-feedstock/pull/47

mgorny:
Yeah, it's coming from:

https://github.com/pytorch/pytorch/blob/3d2e7dee7aa4f7c1e2d1e68b5e7e1c4db1071a86/torch/utils/cpp_extension.py#L1608

[...]

PR #529, opened 2026-08-03
Fix torch.compile on MPS: search the env prefix for c10/metal headers

hmaarrfk:
### Problem

On `osx-arm64`, PyTorch's inductor **MPS** backend JIT-compiles Metal shaders at runtime and embeds the `c10/metal` headers into the shader source — both `torch.mps.compile_shader` and `torch/_inductor/codegen/mps.py` call `_embed_headers`, which resolves `#include <c10/metal/*.h>` by reading the files from `$SP_DIR/torch/include`.

Because this feedstock keeps the `c10` headers in `$PREFIX/include` (per #447) rather than site-packages, `_embed_headers` cannot find them, leaves the `#include` lines untouched, and **every** `torch.compile` targeting the MPS backend then fails at Metal-compile time:

```
fatal error: 'c10/metal/utils.h' file not found
```

Minimal reproducer (osx-arm64, current `pytorch 2.13.0` conda-forge package):

```python
import torch
f = torch.compile(lambda a, b: torch.softmax(a @ b.transpose(-1, -2), -1) @ b)
q = torch.randn(1, 4, 1024, 64, device="mps")
f(q, q)   # InductorError: SyntaxError: failed to compile ... 'c10/metal/utils.h' file not found
```

This silently breaks the entire `torch.compile` → Metal path (FlexAttention on MPS included), not just one op.

### Fix

New patch **`0024`**, a direct companion to **`0006`** (which already redirects the C++ include paths to `$PREFIX/include`): the two Metal-shader `_embed_headers` call sites now also search `sysconfig.get_config_var("prefix") + "/include"`, so the headers are found in the conda layout while the upstream `torch/include` path keeps working for wheels. Recipe-only change, no source symlinks.

Verified locally on osx-arm64: with the patch and no other change, the reproducer above (and `torch.compile` on MPS generally) compiles and runs.

Build number bumped 0 → 1.

### Checklist
- [x] Used a personal fork of the feedstock to propose changes
- [x] Bumped the build number
- [ ] Re-rendered — not needed (recipe-only change; no dependency or `conda_build_config` change)
- [x] Ensured the license file is being packaged (unchanged)


mgorny (approved):
LGTM. We should probably merge the both patches on the next bump.

hmaarrfk:
sorry what 2 patches?

May i ask why we would want to wait on merging this?
I'm happy to cancel every build other apple arm64 since this really only affects that

hmaarrfk:
in fact, i just cancelled the workflows for the others to save on CIs.

Tobias-Fischer (approved):

mgorny:
I meant merge 24 into 6, so we just have the patch that updates all the paths in a single place.

h-vetinari:
> LGTM. We should probably merge the both patches on the next bump.

> May i ask why we would want to wait on merging this?

I feel you're both talking past each other, so allow me to hopefully untangle this; I'm willing to bet that @mgorny suggested to merge the two patches for the next pytorch _version_ bump, not to block this PR on it. So this PR can proceed as-is, and yes, I agree that the patches should be merged when we next have to rebase the lot anyway.

hmaarrfk:
ah thank you both. sorry, "merging" means alot. 

I'll let some1 else click merge, but this seems to be in good shape!

hmaarrfk:
ok i merged the patches together. should be good to go.

hmaarrfk:
Awesome!

mgorny:
Thanks for doing this!

PR #540, opened 2026-09-19
Fix missing sed and broken exit codes for Windows

carterbox:
This PR fixes the Windows build script which is probably broken because it uses `sed`, but `sed` is not installed and the script doesn't correctly raise errors.

Checklist


carterbox:
This first CI run is just a proof of concept to just verify that sed is not available in the CI environment and that the script will raise an error if the syntax is fixed. Non-Windows jobs are skipped because they are irrelevant.

carterbox:
OK. I have proved that SED is missing from the build environment, and error handling was previous not functioning correctly.

```
 Ended : Friday, September 18, 2026 9:11:35 PM

        1 file(s) copied.
'sed' is not recognized as an internal or external command,
operable program or batch file.
  File "D:\Miniforge\Lib\site-packages\conda_build\build.py", line 2549, in build
    windows.build(
```

https://github.com/conda-forge/pytorch-cpu-feedstock/actions/runs/35416830546/job/105827018049?pr=540#step:6:7800

carterbox:
Now on to fixing all the broken things in the script that we didn't know were broken...

carterbox:
I intentionally used an `m2-sed` alternative tool called `sd` because I didn't want to pull in other `m2` dependencies that someone might accidentally depend on in the future without listing them explicitly in the build environment.

carterbox:
We might be affected by this bug: https://github.com/mozilla/sccache/issues/2862

isuruf:
I'd rather use `m2-sed` than `sd`. `m2-sed` is going to be maintained whereas we don't know if `sd` will be. Lots of PRs, but no activity on the repo.

carterbox:
OK.

carterbox:
Need to figure out what to do about these bugs in the latest two sccache versions.

carterbox:
@baszalmstra, I switched to ccache to avoid bugs. Is there a reason that you originally chose sccache? (Some feature that scccache has?)

carterbox:
There is a compile-time bug unique to sm120 (Blackwell) + CCL 2.8.x (CUDA 12.9) + Windows. This bug is not encountered by upstream because they do not publish builds for CUDA 12 + Blackwell.

carterbox:
I didn't realize we were holding Windows back to CUDA 12.8. https://github.com/conda-forge/pytorch-cpu-feedstock/blob/74dd3d46ff0c3d05053197961386b5a2018cf3d1/.ci_support/migrations/cuda129.yaml#L46

